PR #23, opened 2015-06-29
Moving palm oil information under ingredients instead of a menu item

HummingBrid:
I moved the short info about palm oil under ingredients, because that's also how it appears on the website and because palm oil does not justify a menu item by itself.


itchix:
You're right!

PR #131, opened 2016-08-08
Added way to check if an application is installed on a device

Karljoones:
Added the ability to check if an application is installed on a device. Allows the ability to promote the Open Beauty Facts application through the Open Food Facts application only if Open Beauty Facts is not installed on the device. If it is installed, do not promote the application to the user.

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
@@ -1,6 +1,8 @@
 package openfoodfacts.github.scrachx.openfood.utils;
 
 import android.app.Activity;
+import android.content.Context;
+import android.content.pm.PackageManager;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.view.View;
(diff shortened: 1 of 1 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
```diff
@@ -92,4 +94,23 @@ public static Bitmap decodeFile(File f) {
...
+     * @param context the applications context.
+     * @param packageName the package name that you want to check.
+     *
+     * @return true if the application is installed, false otherwise.
+     */
+    private boolean isApplicationInstalled(Context context, String packageName) {
+        PackageManager pm = context.getPackageManager();
+        try {
+            // Check if the package name exists, if exception is thrown, package name does not exist.
+            pm.getPackageInfo(packageName, PackageManager.GET_ACTIVITIES);
```
teolemon: perhaps add org.openbeautyfacts.scanner somewhere ?

Karljoones:
@teolemon I had a look at implementing it in the drawer, have a link to open beauty facts when it is not installed on the device but was unsure on how to without creating a lot of redundant code. The package name would be passed into the method whenever it is called. I guess you just mean add it to the comments so that anyone who is doing it knows how to implement it though. This could be used if there are any more apps in the future!

itchix:
Any news about the implementation? @Karljoones

Karljoones:
@itchix apologies but I have not had a chance (and I partly forgot) to finish this off.

PR #136, opened 2016-09-29
fix Login issue #102

### LINKED ISSUE #102: Login issues
I found two issues in the login process:
* When identifying in the app, if I try with any login (even a random one) and no password, I get a green notification to let me know I'm connected, which makes no sense. Logically, if I give a login, I should give a password as well. 
* After logging in, when I look in the left menu, the login isn't updated (this is the case with a real login / pwd as well). I have to exit with the return button and launch the app again to see the new login in the left drawer.

herau:
use a login activity instead of a login fragment
Update the User profile when the user is connected
Add check on both login and password text fields in the loginView
fix #102


herau:
the build is successful on my local environment.

itchix:
Maybe an issue with the travis file config

PR #139, opened 2016-09-30
Jackson improvements

herau:
refactor all model class to remove useless jackson annotations
Create a Json Utils class to use a shared ObjectMapper ([performance](http://wiki.fasterxml.com/JacksonBestPracticesPerformance))
Optimize json reading
add some unit tests

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/models/FoodAPIRestClientUsage.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/models/FoodAPIRestClientUsage.java
@@ -5,19 +5,20 @@ import android.content.Intent;
 import android.content.SharedPreferences;
 import android.os.Bundle;
 import android.widget.Toast;
+
 import com.afollestad.materialdialogs.MaterialDialog;
-import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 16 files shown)


itchix:
Wow :)

app/src/main/java/openfoodfacts/github/scrachx/openfood/models/FoodAPIRestClientUsage.java
```diff
@@ -277,21 +276,14 @@ public void onStart() {
...
-                        }
-                        onAllergensCallback.onAllergensResponse(true);
-                    } catch (InvalidFormatException e) {
-                        e.printStackTrace();
-                        onAllergensCallback.onAllergensResponse(false);
+                    AllergenRestResponse restResponse = JsonUtils.readFor(AllergenRestResponse.class)
+                            .readValue(responseBody);
+
+                    for (Allergen allergen : restResponse.getAllergens()) {
+                        allergen.save();
```
herau: @itchix do you know what is the goal of this loop on the list to apply a save method on each element ?

itchix: It's to store in the DB all allergens from the API, and then I can use it in this view : AlertUserFragment

herau: ok !

itchix:
Can I pull? =)

PR #153, opened 2016-10-06
code refactoring - lint errors

herau:
make abstract the two activities which are not declared on the manifest file
remove useless parent layout
move some resource types (arrays, styles and strings)
apply some others android & Java lint issues


herau:
local build is OK; don't know what is going on travis

Issue #97, opened 2016-05-22
Nutritional info 


asmoth360:
In the app, the nutritional info are displayed in this order : 
- Salt, Fat, Sugar, Saturated fat

Whereas on the website it's : 
- Fat, Saturated fat, Sugar, Salt

To be consistent and easier to read, it should be the same as on the website.

And the text next to the colored circles is a mix between english and french (for me). By example:
- Sel : moderate
- Matières grasses / Lipides : moderate
- Sucres : low
- Acides gras saturés : high

The terms moderate, low etc. should be translated.

PR #162, opened 2016-10-07
Nutritional info #97

Farbklex:
Fixed https://github.com/openfoodfacts/OpenFoodFacts-androidApp/issues/97.
Changed the order of nutritions in the products nutrition information screen.
Created a method which replaces the nutrition level string from the api with the corresponding localized value.
Added a localisation in German (my native language) and French (copied from the website, so might be not correct in this context).


Farbklex:
It would be nice if you could add the hacktoberfest label here.

Issue #66, opened 2016-04-24
Create list of products by: labels, categories, brands, additives…


teolemon:
- List of : the list of all values that currently exist in Open FoodFacts
- Individual **\* : list of all products with that peculiar value

> 5.6.3 Labels
> 5.6.3.1 List of labels
> 5.6.3.2 Individual label
> 5.6.4 Categories
> 5.6.4.1 List of categories
> 5.6.4.2 Individual category
> 5.6.5 Status of products
> 5.6.5.1 List of States
> 5.6.5.2 Individual Status
> 5.6.6 Packaging
> 5.6.6.1 List of Packagings
> 5.6.6.2 Individual Packaging
> 5.6.7 Brands
> 5.6.7.1 List of Brands
> 5.6.7.2 Individual Brand
> 5.6.8 Purchase Place of products
> 5.6.8.1 List of Purchase Place
> 5.6.8.2 Individual Purchase Place
> 5.6.9 Store of products
> 5.6.9.1 List of Stores
> 5.6.9.2 Individual Store
> 5.6.10 Country of products
> 5.6.10.1 List of Countries
> 5.6.10.2 Individual Country
> 5.6.11 Trace of products
> 5.6.11.1 List of Traces
> 5.6.11.2 Individual Trace
> 5.6.12 Additive of products
> 5.6.12.1 List of Additives
> 5.6.12.2 Individual Additive
> 5.6.13 allergens of products
> 5.6.13.1 List of allergens
> 5.6.13.2 Individual allergen
> 5.6.14 Barcodes
> 5.6.14.1 List of List of barcodes beginning with a given number
> 5.6.14.2 List of barcodes beginning with 3596710
> 5.6.15 Entry dates
> 5.6.15.1 List of entry dates
> 5.6.15.2 List of products with a given entry date
> 5.6.16 Packager codes
> 5.6.16.1 List of Packager codes
> 5.6.16.2 List of products with given Packager code
> 5.6.16.3 List of Packaging cities
> 5.6.16.4 List of products with given Packaging city

teolemon:
Reference at http://en.wiki.openfoodfacts.org/API

herau:
@teolemon in the wiki, this is mention: `(remember to do tests on world.openfoodfacts.net - login and password: off )` But the url require a basic authentication. do you know an openfoodfacts server in which i can do my tests please ?

teolemon:
if it's read, do it on prod.
[...]

PR #182, opened 2016-10-18
[wip] Create list of products by facets

herau:
related to #66


herau:
@itchix
Lots of work here because i switched from loopj to retrofit 2 for all HTTP calls.
I also create the API mappings to list products by facets (ingredients, packaging etc) but it's not yet bind to the view

itchix:
It seems good for me

PR #207, opened 2016-11-07
Bind chrome custom tab to the Sign Up button that's available in the Sign-In view

### LINKED ISSUE #196: Add account creation in Chrome Custom tab
http://world.openfoodfacts.org/cgi/user.pl

Bind it to the Sign Up button that's available in the Sign In view
Bind it to a new entry in the settings area

herau:
*  pre fetching the uri
 * add fallback for device which doesn't support chrome custom tabs
 * copy googleChrome customtabs helper (Apache Licence 2)

fix #196

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -64,7 +64,6 @@ apply plugin: 'com.neenbedankt.android-apt'
 dependencies {
 
     compile 'com.android.support:customtabs:24.0.0+'
-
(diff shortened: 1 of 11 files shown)


herau:
@itchix review time :-)

The current PR has two limitations : 

- Not able to know when the user has done the account creation and then close the Chrome Custom tabs
- The custom icon for the toolbar of the Chrome custom tabs doesn't work. don't really understand why.

PR #321, opened 2016-12-02
More facets

teolemon:
Extended, without quite knowing what I was doing. Hope it will be helpful.


teolemon:
@herau is it useful ? Can it be merged or discarded ?

herau:
@teolemon the JSON responses can't be converted to Search Java class. 

Because the Search class is only for product result : https://github.com/openfoodfacts/OpenFoodFacts-androidApp/blob/master/app/src/main/java/openfoodfacts/github/scrachx/openfood/models/Search.java#L31

You have to create a java class per JSON response to match JSON fields to Java class fields (handling by jackson).

I can update your PR if you want

PR #325, opened 2016-12-04
fix #249 display no photo when no nutrition data

### LINKED ISSUE #249: Weird photo when no nutrition data
screenshot_20161115-130508

naofum:
When simply setting the image size to 1 pixel, image was not displaied but indents the message.


teolemon:
merge ?

PR #328, opened 2016-12-07
Open the product activity after the creation of the product

herau:
refactor open food api client to remove duplicate code
use Async task to add the product in the history
All kind of product accesses are add to the history (search, scan, add)


teolemon:
conflicting

PR #360, opened 2016-12-23
fix #74 The keyboard stays open after Offline edit save

### LINKED ISSUE #74: Better handling of upload problems
[This is a bit of a rant - sorry]

I have installed Open Food 0.1.12 from F-Droid, and was pleasantly surprised when it first found some articles I had and when it then allowed me to anonymously submit info for missing articles.
However, taking pictures was a real PITA (I might make a separate issue report for that); but much more frustrating was that half of the painstakingly taken pictures were lost during upload.

I have added three products. For every product I have added three pictures (main, ingredients, nutrition). Of these nine pictures I now see a total of four pictures at de.openfoodfacts.org; apparently the other five have been lost during upload. That's really frustrating, especially since the entries are gone from my Offline Edit list, so I can't even retry these uploads.

In two cases OpenFood told me about Internet Connectivity issues. In one case this might have been correct since I was a bit far from Wifi hotspot; in another case I took care to be close the hotspot, but the app still complained and lost one picture. In the third case (probably the one where no images were uploaded at all) the app was halted with an error: when it reappeared, one of the products was gone from the Offline Edits lisrt, and none of its images was uploaded.

This makes for a really really frustrating user experience. So I mainly would suggest to keep offline edits in the list and to allow users to retry their uploads. Maybe even allow a re-upload if the upload was successful - while this might be complicated on server-side, I would feel much more in control, rather than having the app just remove my entered data.

naofum:


teolemon:
Thanks @naofum 
I just filed another keyboard related issue at https://github.com/openfoodfacts/OpenFoodFacts-androidApp/issues/369

teolemon:
(This commit is also included in https://github.com/openfoodfacts/OpenFoodFacts-androidApp/pull/370 and https://github.com/openfoodfacts/OpenFoodFacts-androidApp/pull/371)

PR #370, opened 2016-12-24
Fix#78

naofum:


teolemon:
(This commit is also included in #371 )

PR #413, opened 2017-01-08
Fix #254 Add commas between allergenic ingredients

### LINKED ISSUE #254: Add commas between allergenic ingredients
screenshot_20161115-140056

naofum:
screenshot_2017-01-09-00-17-04


herau:
maybe replace by [Chips](https://material.io/guidelines/components/chips.html) soon.

what do you think @teolemon ? maybe also allergens (with different colors according to toxicity)

teolemon:
chips yes, colors: it raises the question of datasource: where does the current one come from ?

herau:
additives are fetched the every time the application is launched (SplashActivity)

teolemon:
I mean: opinions are emitted on additives, but this is not Open Food Facts data.

Issue #412, opened 2017-01-08
Switch from SugarORM to GreenDAO


itchix:
Test in dev-dao branch if Green DAO is better, without bug

herau:
looks like faster and smaller than sugarORM. 
maybe more complex for the schema migration

itchix:
Close with 50d9086a53ade10768d02fd8f637adaff802c724

PR #416, opened 2017-01-08
Use GreenDAO instead of SugarORM

itchix:
ref to #412 @herau


teolemon:
@itchix Have you made some testing ? How will it migrate from the current system ?

app/src/main/java/openfoodfacts/github/scrachx/openfood/views/HistoryScanActivity.java
```diff
@@ -132,7 +133,7 @@ public void exportCSV() {
             }
             String[] headers = {"Barcode", "Name", "Brands"};
             writer.writeNext(headers);
-            List<HistoryProduct> listHistoryProducts = HistoryProduct.listAll(HistoryProduct.class);
+            List<HistoryProduct> listHistoryProducts = Utils.getAppDaoSession(this).getHistoryProductDao().loadAll();
```
herau: maybe use paging to avoid to load the full history

itchix:
It was a mess when I did a rebase, I close and open another PR

Issue #414, opened 2017-01-08
Share icon is pixelated


teolemon:
![image](https://cloud.githubusercontent.com/assets/1689815/21751237/312cb3a8-d5c4-11e6-803e-95321e4ef60e.png)

PR #426, opened 2017-01-10
Fix: Resolve Issue #414 (Share icon is pixelated)

geocohn:
Used internal share icon instead of default
alse replaced png-based share icon with svg-based version


itchix:
Is it possible to get a screenshot of the result ?

geocohn:
screenshot_20170110-143041

PR #496, opened 2017-01-31
remove the LoadToast usage when we are trying to load a product

herau:
to avoid weird blue load toast whereas we trying to apply a Material Design UX.


itchix:
Ok for me =)

teolemon:
It looks like the toast is toast :-D

teolemon:
ref of what's possible in terms of progress indicator in MD: https://material.io/guidelines/components/progress-activity.html#progress-activity-types-of-indicators

itchix:
We need to remove all refs to LoadToast library in the app

PR #531, opened 2017-03-23
Added de translations

fuddl:


herau:
@itchix something wrong with the build but i merge anyway because only contains translations files.

itchix:
yes there is an issue with travis and Java 8 syntax... (Lambda expression)

Issue #324, opened 2016-12-03
Using the locale list from Android instead of our own ?


teolemon:
We try to maintain a locale list in strings.xml. It seems that Android does the job of maintaining a translated list. We could just iterate from strings-xx to see which one we already have and display them using the system API ?
https://developer.android.com/reference/java/util/Locale.html

teolemon:
https://github.com/openfoodfacts/OpenFoodFacts-androidApp/blob/master/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/PreferencesFragment.java

mac-adder:
Changes are ready, one problem though is that the result may not be good enough for some specific languages (see for example this [SO post](http://stackoverflow.com/questions/29536631/getting-locale-variant-for-simplified-chinese)).

The Android core seems to circumvent this by using two dedicated string arrays (one for the special locale codes and the other for their displayable names), but having currently each language displayed using the chosen locale, we can't do the same without missing the original purpose.
Is there any specific reason why this behaviour was chosen in the first place? Is this something we can change?

teolemon:
The current version has the advantage of working with older versions of Android, that might miss some languages. 

The language issue is very complex:

- UX of the application
- Language the food/cosmetic product is available in (the default language displayed, the way to view the product in its other languages)
- Language for new photos/new products (the one of the UX ? Always ask ? Default, but let the user change it )
- Search language (the one of the UX, others ?)

Also, related: 
https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/101

mac-adder:
Actually what I meant was that, in the language list from the preferences, each option is translated based on the current locale, so that if we wanted to use hard-coded names for some specific languages such as simplified chinese, we would have to go back to defining two string arrays per supported locale to remain consistent. 
Apart from avoiding the need for those manual translations, wouldn't it make more sense if each language option was displayed based on its own locale?

teolemon:
@mac-adder @herau Realized this is still an issue in Trunk, and that we had this branch.
Can you turn it into a pull-request ?

PR #542, opened 2017-04-03
Fix #432 (fix handling of languages with compound locales)

### LINKED ISSUE #432: Simplified Chinese and Brazilian Portuguese are not loaded properly
Simplified Chinese and Brazilian Portuguese are not loaded properly while the languages are here. Probably related to the fact that they are not: pt and cn but pt_BR and cn_ZH

mac-adder:


teolemon:
#324 would be a very good candidate to fix. It's slowing translations.

herau:
@itchix @teolemon i will update the code to factorize the locale management

PR #543, opened 2017-04-04
Translations from Crowdin for off, obf, opff and main

teolemon:
This is a new branch generated by Crowdin, a translation system that grabs translations from GitHub, let people translate more easily than Launchpad, and then commits to this branch (l10n_master)
Then we just have to make PR to merge changes.


app/src/main/res/values-es/strings.xml
```diff
@@ -1,26 +1,32 @@
...
   <string-array name="nav_drawer_items_product">
     <item>Resumen</item>
     <item>Ingredientes</item>
     <item>Nutrición</item>
     <item>Tabla Nutricional</item>
   </string-array>
+  <!-- Text units -->
   <string-array name="units_array">
-    <item>g</item>
+    <item> </item>
```
herau: @teolemon sure we want an empty value here ?

teolemon: It looks like it also empty in **values/strings.xml**
Looks like it is to ensure the dropdown is empty by default

teolemon:
@herau Now feeling confident about the merge.

Issue #567, opened 2017-04-15
The import org.greenrobot.greendao.annotation.Unique is never used


teolemon:
```
/Users/pierreslam/Desktop/cloneandroid/openfoodfacts-androidapp/app/src/main/java/openfoodfacts/github/scrachx/openfood/models/HistoryProduct.java
Warning:(6, 8) The import org.greenrobot.greendao.annotation.Unique is never used
```

samagra14:
@teolemon Please close this issue. It has already been resolved.

ross-holloway94:
Issue has been fixed and merged.

Issue #479, opened 2017-01-18
Better looking error message if no results found (or offline)


teolemon:
This is what search results currently look like in offline mode
![image](https://cloud.githubusercontent.com/assets/1689815/22060370/bf72590c-dd70-11e6-8d2a-4e56789f395b.png)

- The reason of no results 'offline' should be displayed along with no results, instead of being shown as a transient message. 
- A try again button should be proposed
- Everything should be centered.
- Cloud optional

 ![image](https://cloud.githubusercontent.com/assets/1689815/22060079/b7f9a85c-dd6f-11e6-83f6-64ceb3ed6390.png)

teolemon:
https://material.io/guidelines/patterns/offline-states.html#offline-states-offline-interactions

teolemon:
![image](https://user-images.githubusercontent.com/1689815/35852602-a5b66cda-0b2b-11e8-85b0-876f4385fb31.png)

alexaldev:
I would like to fix this. Currently, when no results found, no internet connection icon shows up.

huzaifaiftikhar:
@alexaldev This has already been fixed in #1323.

PR #575, opened 2017-04-19
Fixed issue#479

NasiaKouts:


teolemon:
#479

teolemon:
@NasiaKouts The SVG only approach will crash the app for users with older versions of Android. We need to include a raster version.

NasiaKouts:
@herau added new commit, fixing only issue #479

NasiaKouts:
@herau uploaded changes and removed the commit about #567

PR #576, opened 2017-04-19
(import org.greenrobot.greendao.annotation.Unique never used) Fixed issue#567

NasiaKouts:


teolemon:
#567 #479

herau:
@NasiaKouts why we have two another commits link to #479 issue ?

NasiaKouts:
@herau solved only issue #567

PR #593, opened 2017-04-30
Fix for disappearing images >> Update SaveProductOfflineActivity.java

teolemon:
Trying to put more checks to avoid the disappearing front image bug.


teolemon:
@herau first test seemed ok, but still not solved. Will keep tweaking until it works.

Issue #565, opened 2017-04-15
Analyse Linting Results


teolemon:
```
Ran lint on variant offRelease: 1179 issues found
Ran lint on variant obfDebug: 1178 issues found
Ran lint on variant opffRelease: 1184 issues found
Ran lint on variant obfRelease: 1178 issues found
Ran lint on variant offDebug: 1179 issues found
Ran lint on variant opffDebug: 1184 issues found
Wrote HTML report to file:///home/travis/build/openfoodfacts/openfoodfacts-androidapp/app/build/reports/lint-results.html
Wrote XML report to file:///home/travis/build/openfoodfacts/openfoodfacts-androidapp/app/build/reports/lint-results.xml```

Ekt0s:
I've fixed 2 lint errors

samagra14:
@teolemon What exactly needs to be done? Can you please explain in detail? If all these are to be corrected or checked, then what should be the protocol? (Directly submitting a PR or first opening a particular issue followed by a discussion and then a PR ). Please help.

PR #602, opened 2017-05-14
Fixing concatenated string lint error

Ekt0s:
Part of fixing #565 

In this pull request, I've fixed 2 lint error due to string being concatenated from within Java code.

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
@@ -174,7 +174,9 @@ public class Utils {
      *
      * @return true if the application is installed, false otherwise.
      */
+
     public static boolean isApplicationInstalled(Context context, String packageName) {
+    private boolean isApplicationInstalled(Context context, String packageName) {
         PackageManager pm = context.getPackageManager();
         try {
             // Check if the package name exists, if exception is thrown, package name does not exist.


app/src/main/java/openfoodfacts/github/scrachx/openfood/views/SaveProductOfflineActivity.java
```diff
@@ -147,7 +147,7 @@ protected void onCreate(Bundle savedInstanceState) {
         } else {
             mProduct = new SendProduct();
             mProduct.setBarcode(mBarcode);
-			barcodeText.setText(barcodeText.getText() + " " + mBarcode);
+			barcodeText.setText(getString(R.string.format_barcode_text, barcodeText.getText(),  mBarcode));
```
herau: I don't understood the advantage to put the space character in the strings.xml as a string template.
The current code is much more clearer.
what do you think @itchix ?

app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/NutritionInfoProductFragment.java
```diff
@@ -97,7 +97,8 @@ public void onViewCreated(View view, @Nullable Bundle savedInstanceState) {
         List<NutrimentItem> nutrimentItems = new ArrayList<>();
 
         if (isNotBlank(product.getServingSize())) {
-            mTextPerPortion.setText(getString(R.string.nutriment_serving_size) + " " + product.getServingSize());
+//            barcodeText.setText(getString(R.string.format_barcode_text, barcodeText.getText(),  mBarcode));
+            mTextPerPortion.setText(getString(R.string.nutriment_serving_size, Integer.getInteger(product.getServingSize())));
```
herau: Do you want to convert the String value to an Integer value ? Because `Integer.parseInt / Integer.valueOf` is the function to use for that.

teolemon:
@Ekt0s any reason you closed it ?

PR #610, opened 2017-05-31
Finished private rating system

ChasakisD:


ChasakisD:
DummyTestActivity is an activity only for test purposes. We use this to generate a bundle of a product and add it to this activity's intent in order to navigate to RatingFragment from the RatingFragmentTest.

teolemon:
image
image
image
image

Issue #656, opened 2017-12-09
The Storage Permission is asked at the wrong time


teolemon:
The Storage Permission is asked at the wrong time, and thus scary. It is asked when you land on an existing product page.

**Why would I possibly need the storage permission to view a product ?**

- We only need it when **adding a new product from images already on the device**. 
- Otherwise, **if adding a new product from the camera, we only need the camera**.
- **If adding additional photos to an existing product**, we can show the permission request **when the users clicks on the "Add more photos" button**

ross-holloway94:
Fixed in #664

PR #664, opened 2017-12-18
Issue #656

ross-holloway94:
Addressing issue #656 . Storage permission will now be asked for when the user selects "take more pictures".


teolemon:
Thanks, this is a useful one. It was confusing for many. Looks good to me on the feature side of things.
A couple of questions: what happens if of the others buttons is pressed on another page ?
(Add a picture, when no initial picture is present)

ross-holloway94:
> A couple of questions: what happens if of the others buttons is pressed on another page ?
>(Add a picture, when no initial picture is present)

If a picture is present or not, the red floating button only navigates the user back a page. However, when a front picture is not present, there is a "take a picture of the front" option, which just opens the camera.

Issue #621, opened 2017-08-09
0.7.4 App crash when product page is incomplete (drawable-hdpi-v4/error_image.png)


Hangman228:
On Android 4.4.4, app crash when you go on ingredients page who is empty or incomplete, for example on this bar code : 3242272866053 (empty)
or this one : 7613036024358 (incomplete)

Phyks:
Same on Android 4.2.2.

teolemon:
Can you give me the version number of the app, as well as the install source ?

Phyks:
It is version 0.7.4 from [F-Droid](https://f-droid.org/packages/openfoodfacts.github.scrachx.openfood/).

Hangman228:
0.7.4 from Play Store

Hangman228:
Here is a logcat when it crash 

W/ResourceType(25156): Bad XML block: header size 18254 or total size 169478669 is larger than data size 1603
D/AndroidRuntime(25156): Shutting down VM
W/dalvikvm(25156): threadid=1: thread exiting with uncaught exception (group=0x41e7bd88)
E/AndroidRuntime(25156): FATAL EXCEPTION: main
E/AndroidRuntime(25156): Process: openfoodfacts.github.scrachx.openfood, PID: 25156
E/AndroidRuntime(25156): android.content.res.Resources$NotFoundException: File res/drawable-hdpi-v4/error_image.png from xml type xml resource ID #0x7f020064. If the resource you are trying to use is a vector resource, you may be referencing it in an unsupported way. See AppCompatDelegate.setCompatVectorFromResourcesEnabled() for more info.
E/AndroidRuntime(25156): 	at android.content.res.Resources.loadXmlResourceParser(Resources.java:2374)
E/AndroidRuntime(25156): 	at android.content.res.Resources.loadXmlResourceParser(Resources.java:2329)
E/AndroidRuntime(25156): 	at android.content.res.Resources.getXml(Resources.java:996)
E/AndroidRuntime(25156): 	at android.support.graphics.drawable.VectorDrawableCompat.create(Unknown Source)
E/AndroidRuntime(25156): 	at openfoodfacts.github.scrachx.openfood.views.adapters.NutrientLevelListAdapter.getView(Unknown Source)
E/AndroidRuntime(25156): 	at android.widget.AbsListView.obtainView(AbsListView.java:2255)
[...]

PR #665, opened 2017-12-27
Issue fixes

ross-holloway94:
This pull request addresses the issues:
#656 - the permission request has been moved to when "take more pictures" is selected.
#621 - The error image has been replaced by a vector drawable, as the program was expecting.


teolemon:
- #621 : We need a vector drawable, but also a raster fallback
- Please keep pull request separated for different issues, so that they can be reviewed more easily :)

ross-holloway94:
Actually, my mistake. I probably shouldn't do a pull request from my master branch. Sorry.

Issue #666, opened 2017-12-31
Add Allergen button doesn't work


teolemon:
![image](https://user-images.githubusercontent.com/1689815/34463241-388964b6-ee55-11e7-9a17-9d274854bc24.png)
There should be a white plus icon within the button

ross-holloway94:
Fixed in #670

PR #670, opened 2018-01-01
Fix 666 add allergen button

ross-holloway94:
Addresses #666 . Button now moved to bottom right of screen and "+" displayed in white.
2018-01-01 16_25_35-android emulator - nexus_5x_api_26_5554


teolemon:
Looks good to me.

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
```diff
@@ -176,7 +176,7 @@ public static Bitmap decodeFile(File f) {
      */
 
     public static boolean isApplicationInstalled(Context context, String packageName) {
-    private boolean isApplicationInstalled(Context context, String packageName) {
+    //private boolean isApplicationInstalled(Context context, String packageName) {
```
alikyo: Why changing the visibility to public ? is it used out side ?
if yes could you remove the commented code please.

teolemon: It was commented because it was crashing the app. Not sure how the duplicate got there in the first place.

teolemon:
@alikyo @ross-holloway94 I'm confused by the locales.
https://developer.android.com/reference/java/util/Locale.html

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
ross-holloway94: Yeah, I'm not sure. The duplicate is preventing building.

ross-holloway94:
@teolemon Locale was also preventing build. All I did here was delete "sr_rCS" and "sr_CS" which are both invalid names, but both seemed to me like duplicates.

teolemon:
Pulling from master should have pulled the private ratings branch, which I hastily merged. And that should break everything because of a missing ORM update.

Issue #680, opened 2018-01-26
Strings not resources in activity_save_product_offline.xml


Karljoones:
Strings should be resources to ensure translation

ross-holloway94:
Fixed in #683

Issue #343, opened 2016-12-20
Reorder Ingredients and Nutrition in the dropdown menu for new product


teolemon:
- Front
- Ingredients
- Nutrition

It's that way in V1, in the Web view, and it's mentally disturbing for regular contributors

teolemon:
Also please display it more like this:
![image](https://cloud.githubusercontent.com/assets/1689815/21385172/d2be8936-c76d-11e6-9e70-b84f5262cc92.png)

ross-holloway94:
What is the name of the layout file to change this? I cannot seem to find it.

teolemon:
activity_save_product_offline.xml

ross-holloway94:
Fixed in #683

PR #683, opened 2018-01-27
Move Ingredients above Nutrition for a new product.

ross-holloway94:
#343 (unintentionally addresses #680 too)


teolemon:
Merged, thanks @ross-holloway94 :-)

Issue #677, opened 2018-01-18
The categories view stays stuck on "Loading…"


teolemon:
```

01-18 15:34:58.233 2005-2030/openfoodfacts.github.scrachx.openfood D/OpenGLRenderer: endAllActiveAnimators on 0x7426b5bc00 (StateListDrawable) with handle 0x74362ea720
01-18 15:35:08.362 2005-12890/openfoodfacts.github.scrachx.openfood E/openfoodfacts.github.scrachx.openfood.views.viewmodel.category.CategoryFragmentViewModel: Error loading categories
java.net.SocketTimeoutException: timeout
at okio.Okio$4.newTimeoutException(Okio.java:230)
at okio.AsyncTimeout.exit(AsyncTimeout.java:285)
at okio.AsyncTimeout$2.read(AsyncTimeout.java:241)
at okio.RealBufferedSource.indexOf(RealBufferedSource.java:345)
at okio.RealBufferedSource.readUtf8LineStrict(RealBufferedSource.java:217)
at okio.RealBufferedSource.readUtf8LineStrict(RealBufferedSource.java:211)
at okhttp3.internal.http1.Http1Codec.readResponseHeaders(Http1Codec.java:187)
at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.java:88)
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:45)
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:93)
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.java:93)
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:125)
[...]

PR #696, opened 2018-02-09
Issue 677

policorn:
This PR fix Issue #677 by increasing read timeouts from 10 to 30 seconds. The categories client does not use the OFF Api client so I've increased both client's timeouts just in case (as other functionalities pulling big JSON streams from the DB might fail the same way)


teolemon:
Perfect. Merged.

Issue #716, opened 2018-02-12
Adding PULL_REQUEST_TEMPLATE.md file for new contributors


iml-v:
There should be PULL_REQUEST_TEMPLATE.md file which tell new contributors how to get started for contribution .

sachincool:
I could Suggest some New Changes in the existing Template, Can I work on this? @iml-v @teolemon

PR #718, opened 2018-02-12
Created PULL_REQUEST_TEMPLATE.md for new contributor

iml-v:
fix issue :#716  created PULL_REQUEST_TEMPLATE.md  with some instruction(what to do ) before sending PR ,for new contributor in project .

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,20 @@
+## Description
+
+Describe the changes made and why they were made instead of how they were made.
+
+## Related issues and discussion
+#{Issue Number}
+ 
+ ## Screen-shots, if any
+ 
+ ## Checklist
+ 
+ Please make sure these boxes are checked before submitting your pull request - thanks!
+ 
+ 
+ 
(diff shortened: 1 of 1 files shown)


iml-v:
@teolemon review this PR.

teolemon:
@ross-holloway94 what do you think of it ?

ross-holloway94:
@teolemon Mostly I agree with what is stated. I disagree with squashing the commits though - that will prevent you cherry-picking or reverting.

iml-v:
@ross-holloway94  if someone have made many commits for the single task ,then it is good that he squash them  and make single commit ,for different issues it is not recommended , it's only when multiple commit for single issue .i would like to hear from you on this .

ross-holloway94:
@iml-v In my opinion, issues should be separated into branches, and then commits can be left separate on that particular branch. I believe this is how @teolemon prefers it?

iml-v:
@ross-holloway94 that is right (separated branches) but my point is that when we do multiple commit regarding single task then before sending PR squash them in single so the whole work can come in single commit .so now should i make changes in Template ? @ross-holloway94  @teolemon  .

Issue #486, opened 2017-01-21
Harmonize color and size of the share icon in product page


teolemon:
Small details about sharing. See https://github.com/openfoodfacts/OpenFoodFacts-androidApp/issues/185 for details on how to implement native sharing.

![image](https://cloud.githubusercontent.com/assets/1689815/22174023/d7426f6c-dfd3-11e6-91a2-198557a775f3.png)

SatyamBansal:
I want to work on this issue

Karljoones:
Fixed #779

PR #768, opened 2018-02-15
Fixes #519 Share icon is grey while edit icon is black

### LINKED ISSUE #519: Share icon is grey while edit icon is black
On product view, the Share icon is grey while edit icon is black. They should both be black and the same size.

ManavPoddar:
## Description

Did changes in the styles.xml file. This was done to create rules for the action menu items for the preffered style. 

Now the share button is also black. 
 ## Screen-shots, if any
 
screenshot_20180216-021535

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/FullScreenImage.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/FullScreenImage.java
@@ -3,7 +3,6 @@ package openfoodfacts.github.scrachx.openfood.views;
 import android.content.Intent;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
-import android.widget.Toast;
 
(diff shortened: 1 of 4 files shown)


Karljoones:
Are these icons different sizes? Maybe you could throw a commit into this pull request that fixes the sizing issues? Unless I am crazy and they are the same size, if so, ignore me

policorn:
Hi @Karljoones the sizes don't change, I think it's a optical illusion. I've just checked out this PR and compared with the master: 
gray_and_black

@teolemon I think this can be merged and the issue #519 can be closed

teolemon:
The share icon is larger that the edit in master as well ? Shouldn't they be the same size ?

Karljoones:
The size difference is a different issue, but it should just be combined into a single PR :)

teolemon:
#486

Issue #764, opened 2018-02-15
There is an empty list item in the start of the categories tab.


ManavPoddar:
**Summary**
The list item in the start of categories is empty

**Steps to reproduce:** 
menu->categories

**Screen-shots:** 

![screenshot_20180215-214337](https://user-images.githubusercontent.com/18038492/36267517-143793c6-129a-11e8-93d0-7919bf3a2707.png)

Shubham2320:
I would love to work on this issue if available.

ManavPoddar:
I am already working on it

PR #790, opened 2018-02-17
Issue #764 There is an empty list item in the start of the categories tab.

ManavPoddar:
## Description

Since there was an emty item in the categories list, so simply deleted the first element.

 ## Screen-shots, if any
 
screenshot_20180215-214337
screenshot_20180218-003027


ManavPoddar:
@Karljoones ok, i'll update the PR after doing the required changes.

One more thing, I have used a separate branch for this pull request, but then also my other commits are being shown along with this commit. Any idea what i'm doing wrong ?

Karljoones:
@ManavPoddar unfortunately, I am not really the best person to ask about this. Might be best to ping the Slack group about this to get their feedback

Karljoones:
Hi @ManavPoddar I'm going to close this due to another PR being opened with #795 :)

PR #795, opened 2018-02-18
Fixes #764 There is an empty list item in the categories tab

### LINKED ISSUE #764: There is an empty list item in the start of the categories tab.

ManavPoddar:
## Description

Added a check for checking if the item is empty or not. if empty then the item is removed

 
 ## Screen-shots, if any
 screenshot_20180215-214337

screenshot_20180218-003027


ManavPoddar:
The travis CI build is failing although the app is running perfectly fine. Any idea what is wrong ?

app/src/main/java/openfoodfacts/github/scrachx/openfood/category/mapper/CategoryMapper.java
```diff
@@ -28,6 +28,10 @@ public CategoryMapper() {
                     tag.getProducts()));
         }
         Collections.sort(categories, (first, second) -> first.getName().compareTo(second.getName()));
+        if(categories.get(0).getName().isEmpty()||categories.get(0).getId().isEmpty()){
```
policorn: Hi @ManavPoddar, what would happen if the categories list is empty? Can you also write some unit tests covering those scenarios?

ManavPoddar: Hi @policorn , Yeah i can do that but i think the above method wont work, as it is not being passed by travis CI build. Maybe have to find another way around, but i'll surely try :)

Karljoones:
How is this coming along @ManavPoddar? :)

ManavPoddar:
Hi @Karljoones, I've added an empty view to the categories_list_fragment, which would be displayed when there is no result or internet is not working and the categorie list doesn't load up. Right now I'm trying to figure out how to access the categoriesList from the fragment to put the condition. :)

Karljoones:
Awesome, thanks @ManavPoddar! Keep up the good work :)

teolemon:
What's up on this one ?

ManavPoddar:
hi @teolemon 
The problem i'm facing with this issue is that the categoriesList take too much time to load in the categories page. So the list is  empty before the list completely loads. 

Thus, I used getViewTreeObserver().addOnGlobalLayoutListener to check the status of the list at every point of time. So when the list completely loads, and if it is empty then we can show the empty view or else we could continue with the normal view.

But the problem in this method is that since the Listener checks the status of the list at every instant, The    
empty view is displayed even during the loading time.





getviewtreeobserver
runcheck

  


One more way I thought of was to check for the empty category list in the category mapper class. Then using this check in the categoryListFragment and display the respective views. But for this method we have to use shared preferences but i am unable to implement shared preferences in the category mapper class.

PrajwalM2212:
@ManavPoddar  Update thia pull request   . This pull request was opened 2 weeks ago .

ManavPoddar:
Hey guys, I'm closing this PR and will be opening a new PR on this issue.

PR #808, opened 2018-02-19
fixes #805 Persistent dismissal of image upload message

### LINKED ISSUE #805: Persistent dismissal of image upload message
**Summary:** 

The message that informs the user that only one image is necessary for an upload should not be displayed once the user dismisses this. You can check out the message that was added in #789. The message should always appear, but once dismissed by the user, this should be stored in prefs and never shown again.

**Expected behavior:** 

This should always display unless dismissed by the user. Once dismissed by the user, it should never show again.

Shubham2320:
## Description

SharedPreference is used to store the boolean (message is dismissed or not).

fixes #805

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/SaveProductOfflineActivity.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/SaveProductOfflineActivity.java
@@ -58,26 +58,41 @@ public class SaveProductOfflineActivity extends BaseActivity {
     private final String[] mUnit = new String[1];
     @BindView(R.id.toolbar)
     Toolbar toolbar;
-    @BindView(R.id.barcodeDoubleCheck) TextView barcodeText;
-    @BindView(R.id.imageSaveFront) ImageView imgSaveFront;
-    @BindView(R.id.imageSaveNutrition) ImageView imgSaveNutrition;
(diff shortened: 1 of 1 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/views/SaveProductOfflineActivity.java
```diff
@@ -88,7 +103,11 @@ protected void onCreate(Bundle savedInstanceState) {
         getSupportActionBar().setDisplayHomeAsUpEnabled(true);
 
         mSendProductDao = Utils.getAppDaoSession(this).getSendProductDao();
-
+        mSharedPref = getApplicationContext().getSharedPreferences("prefs", 0);
+        boolean messageDismissed = mSharedPref.getBoolean("is_message_dismissed", false);
```
teolemon: Can we be much more explicit about the message ? I expect that eventually we'll have many explainer messages at many places in the app.

Shubham2320:
@teolemon I've made the changes please review :).

teolemon:
Is a migration script necessary for the ORM ? @itchix ?

Shubham2320:
Are there any more changes which I need to do?

itchix:
@teolemon I don't think so
He use `getSharedPreferences` so there is no modification in the ORM

teolemon:
perfect, let's merge. thanks @Shubham2320

Issue #793, opened 2018-02-17
Forever loading on SplashScreen on OffDebug Variant 


sachincool:
**Summary:** 
The Splash Screen Keeps on Loading on `OffDebug` Variant 

**Steps to reproduce:** 
1 choose offDebug as your build variant
2 Build and run the variant on 7.1 Android 
3 Splash Screen with LoadToast is showing indeterminately. 

**Add System logs:**
There's no Error 
Only a bunch of Database create calls. 
**Expected behavior:** 
The App should run fine on android 7.1

**Observed behavior:** 
I checked further and got that it Leads to GetJson Asynktask to execute and get the json value. 
Afterwards it runs the preexecute and Doinbackground Perfectly but gets stucked on PostExecute. 
![device-2018-02-18-045801](https://user-images.githubusercontent.com/25671488/36346688-44a0b7d8-1469-11e8-84ca-e039bdc10901.png)
![1](https://user-images.githubusercontent.com/25671488/36346687-44559e92-1469-11e8-872d-a4b2471bfdb0.png)
![screenshot from 2018-02-18 04-59-17](https://user-images.githubusercontent.com/25671488/36346689-44e6ec6c-1469-11e8-830a-98c8fccf9d49.png)

**Device and Android version:** 
Android 7.1
*Yes, I would like to work on this issue*

sachincool:
Yes i would like to Work on this issue.

sachincool:
It is caused inside that `getAllergence` call And inside that call it calls the Overridden method `onFailure` 
Here is the proof:
![1](https://user-images.githubusercontent.com/25671488/36346770-34daae06-146b-11e8-934c-8b0c82e94178.png)

sachincool:
on printing the stacktrace of the throwable `t`  i got this 
```
02-18 05:32:02.130 24333-24333/openfoodfacts.github.scrachx.openfood W/System.err: java.net.SocketException: Network is unreachable
02-18 05:32:02.131 24333-24333/openfoodfacts.github.scrachx.openfood W/System.err:     at java.net.PlainSocketImpl.socketConnect(Native Method)
        at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:334)
        at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:196)
        at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:178)
[...]

PR #812, opened 2018-02-19
Fix Freeze on Nougat

sachincool:
## Description

The `offdebug`  variant no longer  freezes at SplashScreen on nougat.

https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/793
 
 ## Screen-shots, if any
 The build no longer freezes on nougat.
 ## Checklist
 * I think it was a general problem and not only for nougat devices But for the offDebug. 
 * i need some guidance, this was a fix only capable because `getAllergence` Method fails.  I need some help understanding why do we need this method and why is it only getting called in `offDebug`. But now The app opens without any crash and acts normally like `obfDebug` build.
 
 
 - [x] Read and understood the contribution guidelines .


policorn:
HI @sachincool, were you able to reproduce #793 in en emulator?

sachincool:
28308410_1508373292593157_1983838524_n

Yep @policorn  This is the old code Pic 
And after the line add i can access the MainActivity, But I haven't seen a  nougat 7.0 specific code inside the app So i am a bit doubtful about this fix. i want to know about the `getAllergens()` Method in OpenFoodApiClient.java.

app/src/main/java/openfoodfacts/github/scrachx/openfood/network/OpenFoodAPIClient.java
```diff
@@ -304,7 +304,7 @@ public void onResponse(Call<AllergenRestResponse> call, Response<AllergenRestRes
 
             @Override
             public void onFailure(Call<AllergenRestResponse> call, Throwable t) {
-
+                onAllergensCallback.onAllergensResponse(false);
```
policorn: Indeed it allows the user to reach the MainActivity in case of network error downloading `allergens.json` from the server on start

sachincool:
@policorn  I'll Add the Final Fix With this change combined, This has alot of priority I'll try to do it today asap.  
Closing this one.

PR #894, opened 2018-02-24
Fix text size

PrajwalM2212:
## Description

The Intro Screens are awesome but text sizes must be in sp .

#895


policorn:
Thanks @PrajwalM2212! Merged

PR #931, opened 2018-02-26
fixes #930 In AlertUserFragment we're only abstracting english and  french allergens from AllergenDao

### LINKED ISSUE #930: In AlertUserFragment we're only abstracting english and french allergens from AllergenDao
**Code:**

```java
for (Allergen a : mAllergensFromDao) {
            if (Locale.getDefault().getLanguage().contains("fr")){
                if(a.getIdAllergen().contains("fr:")) allS.put(index, a.getName().substring(a.getName().indexOf(":")+1));
            } else if (Locale.getDefault().getLanguage().contains("en")) {
                if(a.getIdAllergen().contains("en:")) allS.put(index, a.getName().substring(a.getName().indexOf(":")+1));
            }
            index++;
        }
```
Here in this code, we're only abstracting English and French allergens. However, I've checked allergens of other languages are also present in AllergenDao so instead of just these two languages we should check for each language which has been chosen by the user.

I've already fixed this and going to send a PR soon.

Shubham2320:
## Description

Fixes #930 and also optimised the code little bit as in previous code `Locale.getDefault().getLanguage()` was called for each element of  AllergenDao so instead of this it is now called just one time before starting of the loop.

**Example of languages which are in AllergenDao and are not used before:**
Italian, German, Japanese, Russian, etc.

#930

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AlertUserFragment.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AlertUserFragment.java
@@ -82,17 +82,15 @@ public class AlertUserFragment extends BaseFragment {
 
     @OnClick(R.id.fab)
     protected void onAddAllergens() {
-        final LinkedHashMap<Integer,String> allS = new LinkedHashMap<>();
+        final LinkedHashMap<Integer, String> allS = new LinkedHashMap<>();
         int index = 0;
+        String language = Locale.getDefault().getLanguage();
         for (Allergen a : mAllergensFromDao) {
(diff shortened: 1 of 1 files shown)


teolemon:
This makes me think that we should look for all translations of an allergen, in all available languages on the product. This might yield false positives but will be safer than the current implementation.

Shubham2320:
yep, I agree.

teolemon:
merging, we need to open an issue about my remark

Shubham2320:
Shall I open a new issue?

teolemon:
Please do, you can copy paste directly from here

PR #942, opened 2018-02-26
Scan first product

PrajwalM2212:
## Description

If the history is empty, a "Scan first product" button is added.

#166 
 
 ## Screen-shots, if any
scanfirst


Shubham2320:
I think one image or a textView with the same message should also appear. At present, the layout seems quite empty.

PrajwalM2212:
@Shubham2320  The issue wanted only a button . Let us see what @teolemon has to say .

Shubham2320:
yeah, that can be added later. :)

teolemon:
Feel free:
-to ask for clarifications before starting coding
- to be creative, and do something smarter than my boring suggestions :-)

More seriously, this is a good start, but it indeed looks a bit empty. I guess we could use the blue and white button design we use in the rest of the app.

We could also add a descriptive text above the button like: "All the products you scan or search for will appear here. Swipe left if you want to remove a product. Your history is yours: it is never sent online, and you can easily export it to use it outside Open Food Facts. "

jaztriumph:
I think we should take away the 'no data' toast and manage to show in a text view itself because I think toast don't look good.

PrajwalM2212:
@jaztriumph  Thanks for the suggestion .

PrajwalM2212:
empty_history

Would this be okay ?

teolemon:
Can you remove the yellow background make the text gray ? For the "Scan first product" can you reuse the same style as the Remove allergen buttons in the allergen view ?

Karljoones:
I would second the changes that @teolemon mentioned, however, it might be best to change the text. Maybe remove the ALL CAPS? And rephrase:

> Your viewed product history will be listed here! 
> This history is for your eyes only and is stored locally.

PrajwalM2212:
I will do the changes.

PrajwalM2212:
scan

PrajwalM2212:
@teolemon @Karljoones Made the changes.

Karljoones:
@prajwalm2212 much better! A couple of final things though. Can you use center justification with the text? And maybe a little bit more padding between the two and finally, the text just needs to be a little greyer to more blend into the background!

teolemon:
Also punctuation at the end of sentences. :)

PrajwalM2212:
scan
@Karljoones  The text was already center justified . I have added more padding . I have made the text greyer as you requested .
@teolemon I have added punctuations at the end of sentences.

Karljoones:
Hi, the text is centered but the justification of the text is not. Off the top of my head, if you use android:gravity="center" might resolve it.

PrajwalM2212:
scan
@Karljoones  I think this is what you were looking for.

app/src/main/java/openfoodfacts/github/scrachx/openfood/views/HistoryScanActivity.java
```diff
@@ -290,4 +312,39 @@ public void onDraw(Canvas c, RecyclerView parent, RecyclerView.State state) {
...
+                Intent intent = new Intent(HistoryScanActivity.this, ScannerFragmentActivity.class);
+                startActivity(intent);
+            }
+        }
+
+    }
+
+    public void setInfo(TextView view) {
+
+        String info = "     Your viewed product history will be listed here.\n" +
```
teolemon: Are the spaces normal ?

PrajwalM2212: @teolemon To get the text centered justified , I had to put some spaces . I had already used android:gravity="center" but that did not solve the issue .

PrajwalM2212: @teolemon  I have solved it now . Thanks to suggestion from @Karljoones

PrajwalM2212:
@teolemon 
It is now centre justified on smaller devices also .
scan

Karljoones:
Thanks for making the changes @PrajwalM2212!

PR #959, opened 2018-02-27
fixes #877 Additives are always displayed in English while we have tr…

### LINKED ISSUE #877: Additives are always displayed in English while we have translations
https://world.openfoodfacts.org/api/v0/product/737628064502.json

https://world-fr.openfoodfacts.org/additives.json
https://world-ru.openfoodfacts.org/additives.json
https://world.openfoodfacts.org/additives.json

image

Shubham2320:
## Description

On `PreferanceChangeListener` of **PreferencesFragment**, I've called a method `GetAdditives` which uses AsyncTask to get additives in the selected language and stores it in the **greenDao** database so that next time when a user visits **IngredientsProductFragment** he/she can see additives in the selected language.

At present only English and French translation can be seen.

#877

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/PreferencesFragment.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/PreferencesFragment.java
@@ -1,29 +1,58 @@
 package openfoodfacts.github.scrachx.openfood.fragments;
 
+import android.content.Context;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.content.res.Configuration;
+import android.net.ConnectivityManager;
+import android.net.NetworkInfo;
 import android.net.Uri;
+import android.os.AsyncTask;
 import android.os.Build;
 import android.os.Bundle;
 import android.support.v4.app.FragmentActivity;
+import android.support.v4.content.ContextCompat;
 import android.support.v7.preference.ListPreference;
 import android.support.v7.preference.Preference;
 import android.support.v7.preference.PreferenceFragmentCompat;
+import android.util.Log;
 import android.widget.Toast;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+
+import net.steamcrafted.loadtoast.LoadToast;
 
 import org.apache.commons.text.WordUtils;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
 import java.util.Locale;
 
 import openfoodfacts.github.scrachx.openfood.R;
+import openfoodfacts.github.scrachx.openfood.models.Additive;
+import openfoodfacts.github.scrachx.openfood.models.AdditiveDao;
+import openfoodfacts.github.scrachx.openfood.network.OpenFoodAPIClient;
+import openfoodfacts.github.scrachx.openfood.utils.JsonUtils;
 import openfoodfacts.github.scrachx.openfood.utils.LocaleHelper;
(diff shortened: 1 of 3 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/PreferencesFragment.java
```diff
@@ -62,14 +90,80 @@ public boolean onPreferenceClick(Preference preference) {
...
+        }
+
+        @Override
+        protected Boolean doInBackground(Void... arg0) {
+            boolean result = true;
+            Context context = getActivity();
+
+            String additivesFile;
+            switch (LocaleHelper.getLanguage(context)) {
+                case "fr":
```
teolemon: Either doing a very long case system with all the cases, or concatenate languages, and handling languages with non standard codes.

Shubham2320: yeah, I thought an approach before to do something like this:
```java
String language = LocaleHelper.getLanguage(SplashActivity.this);
            String additivesFile = "additives_en.json";
            additivesFile.replace("en", language);
```
but I'm unsure about having translation in every language in server at present.

Shubham2320: But if the user has changed the language and the selected language doesn't contain the translation so in that case, additives would not be changed to the new language and remain as it is.
It seems quite approachable by the code I've shown.

Shubham2320:
@teolemon I've made the changes please review.

app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/PreferencesFragment.java
```diff
@@ -62,14 +90,67 @@ public boolean onPreferenceClick(Preference preference) {
...
+            lt.show();
+        }
+
+        @Override
+        protected Boolean doInBackground(Void... arg0) {
+            boolean result = true;
+
+            String language = LocaleHelper.getLanguage(getActivity());
+            String additivesFile = "additives_en.json";
+            additivesFile.replace("en", language);
```
Karljoones: Why not replace with these three lines with one?
`String additivesFile = "additives_" + LocaleHelper.getLanguage(getActivity()) + ".json";`

app/src/main/java/openfoodfacts/github/scrachx/openfood/views/SplashActivity.java
```diff
@@ -105,21 +100,15 @@ protected Boolean doInBackground(Void... arg0) {
...
-                    additivesFile = "additives_fr.json";
-                    break;
-                case "en":
-                default:
-                    additivesFile = "additives_en.json";
-                    break;
-            }
+            String language = LocaleHelper.getLanguage(SplashActivity.this);
+            String additivesFile = "additives_en.json";
+            additivesFile.replace("en", language);
```
Karljoones: Same as previous comment.

Shubham2320:
I've made the changes and tested it by switching from English to French in this case the app works fine. I'm not sure about other languages but I've found in API that translations in other languages are also there. It'll take some more time to test it completely :).

Karljoones:
Merged, thanks @Shubham2320 :)

Shubham2320:
Thanks for merging my code :).
However, some more changes would be required in it in future to allow functionality in every language.

PR #988, opened 2018-02-28
fixes #951 Add a warning about allergen detector when a product is not complete

### LINKED ISSUE #951: Add a warning about allergen detector when a product is not complete
When the user has set at least an allergen to detect, add a warning about allergen detector when a product is not complete.

"This product is not complete. As a result, we couldn't do the allergen check."

Shubham2320:
## Description

To get the array `states_tags` from the server I've added `states_tag` in `getFullProductByBarcode()` method's relative URL in `OpenFoodApiService` and used this to check product is complete or not in `SummaryProductFragment` if the product is incomplete a dismissable message is displayed in the UI.

fixes #951 
 
 ## Screen-shots, if any


Karljoones:
Thank for the work @Shubham2320, can you ensure that the layout for this matches the message regarding only one picture being necessary? I don't have the code right now to show you but I hope you understand.

app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/SummaryProductFragment.java
```diff
@@ -141,6 +142,11 @@ public void onViewCreated(View view, @Nullable Bundle savedInstanceState) {
         final State state = (State) intent.getExtras().getSerializable("state");
 
         final Product product = state.getProduct();
+
```
teolemon: can you add a check to display it if the user has at least one allergen in their Alert list ?

Shubham2320: Sure :).

Shubham2320:
## Screenshot:



@Karljoones @teolemon  I've made some minor changes and now the two layouts are same :) also I've added the code to check if the user has at least one allergen in their Alert list before showing the message please review :).

Karljoones:
I have reviewed the code and can confirm that it is the exact same as the offline product entry notification so there is absolutely nothing wrong with the code. However, I do not think that this blends into this activity, maybe it's because the rest of the text is grey compared to the back. I would like @teolemon to chip in on this one to be honest. If he is happy with it, then I am happy to merge

teolemon:
@Karljoones this is a safety fix. We can make it more beautiful later.

Karljoones:
Merged, thanks @Shubham2320 :)

PR #979, opened 2018-02-27
Show grade nutrition on search and history view

itchix:
## Description

It's been a long time since I made a commit 😋 
So this PR to add some information about grade nutrition when somebody want to search a product.
I did the same add for history.

## Related issues and discussion
None
 
 ## Screen-shots, if any
screenshot_20180227-223503
screenshot_20180227-224219
 

 ## Checklist
 
Make sure you've done all the following (You can delete the checklist before submitting)
 
 - [x] Apply the `AndroidStyle.xml` style template to your code in Android Studio.
 - [x] Code is well documented
 - [ ] Include unit tests for new functionality
 - [x] All user-visible strings are made translatable
 - [x] Code passes Travis builds in your branch
 - [x] If you have multiple commits please combine them into one commit by squashing them.
 - [x] Read and understood the contribution guidelines .

Issue #925, opened 2018-02-25
Make the history layout more consistant with the search layout


teolemon:
Current search: https://user-images.githubusercontent.com/1689815/36647480-913da272-1a86-11e8-8cdd-49973013ec46.png

Proposed history:
![image](https://user-images.githubusercontent.com/1689815/36647494-c90f9926-1a86-11e8-9485-c1e815020f63.png)

Karljoones:
Related: #840

Shubham2320:
Seems good. Can I work on it?

l-o-b-s-t-e-r:
Looks like there is no progress here, so I've taken this issue.

PR #1000, opened 2018-03-01
Updated History and Product search UI

l-o-b-s-t-e-r:
- created small icons for nutri score
 - changed UI of History Product
 - changed Entity of History Product
 - implemented Database migration
 - REMOVED loading of FULL size images, instead of this I added loading via Picasso (this small fix 
speeded up displaying of products several times)


l-o-b-s-t-e-r:
#925

l-o-b-s-t-e-r:
I've spent like an hour to understand why history screen is loading so looong. It was really strange, because we just get data from the local DB. Finally, I've found that we were loading the full-size image of the each product...only after loading, history products were shown.
I even cannot imagine how long users are waiting if they have more than 100 products...

l-o-b-s-t-e-r:
@teolemon 
@Karljoones 
@policorn

l-o-b-s-t-e-r:
alt text

PrajwalM2212:
Looks Awesome !
And great work identifying the cause of delayed load .

itchix:
You can put the nutriscore grade on search view too, like this PR : https://github.com/openfoodfacts/openfoodfacts-androidapp/pull/979

I'll will close mine if you do it :smiley:

l-o-b-s-t-e-r:
@itchix no problem :)

itchix:
Please fix travis issue : 
```
Caused by: java.lang.RuntimeException: Constructor (see HistoryProduct:35) has been changed after generation.
Please either mark it with @Keep annotation instead of @Generated to keep it untouched,
or use @Generated (without hash) to allow to replace it.
```

l-o-b-s-t-e-r:
@itchix 
Fixed)

policorn:
Let's wait for quorum and merge : )

l-o-b-s-t-e-r:
@Karljoones I hope no, but I'm ready to protect my code 😎🔫😂

itchix:
Ahaha it's OK for me!
I'll close my PR

jaztriumph:
@l-o-b-s-t-e-r I think putting a progress bar on the history image holder so that user knows when image is loading will be nice. It will be just 2 lines of code to do with picasso callback function.
progress

teolemon:
Ok, merging .Thanks  @l-o-b-s-t-e-r 
The loader thing would indeed be nice to have @jaztriumph

jaztriumph:
@teolemon then I will make a issue and implement it.

PR #1011, opened 2018-03-01
Message shown on empty Username.

danzzen:
## Description
Message will be shown First for Username when user will try to sign in without filling data for Username and Password. Previously when username and Password field were empty and user tries to sign in error message was for the password field ,not for the Username field.
 
 ## Screen-shots
<h4>Previously</h4>

whatsapp image 2018-03-01 at 7 21 48 pm 1

<h4>Expected</h4>

whatsapp image 2018-03-01 at 7 21 48 pm


Karljoones:
Marged, thanks @danzzen!

PR #1026, opened 2018-03-01
 java.lang.NullPointerException and trim extra space 

danzzen:
## Description

` getActionBar().setDisplayHomeAsUpEnabled(true);` can produce  java.lang.NullPointerException so this warning is fixed and extra `space` at the end of text is trimmed.


policorn:
Tested. Approved

app/src/main/java/openfoodfacts/github/scrachx/openfood/views/LoginActivity.java
```diff
@@ -92,22 +93,19 @@ protected void onCreate(Bundle savedInstanceState) {
 
     @OnClick(R.id.buttonSave)
     protected void attemptLogin() {
-        String login = loginView.getText().toString();
-        String password = passwordView.getText().toString();
-
+        String login = loginView.getText().toString().trim();
+        String password = passwordView.getText().toString().trim();
```
policorn: Not sure if OFF password policy allows 'space' as a valid character. The common sense tells me it should be allowed and not trimmed as it's a valid ASCII character

policorn: I have checked the server-side code; The only limitation is the password must have at least 6 characters, so It's confirmed space is a valid character and should not be removed

Karljoones: @policorn should we close this PR as it is not something we would like to implement in case it breaks log in?

danzzen: @policorn yes you are right space will be valid for password i will make changes.Thanks!

Karljoones:
I'm going to close this as you have created a new PR #1032 :)

PR #1032, opened 2018-03-02
Closes #996  Login issue fixed

### LINKED ISSUE #996: Already Login is displayed when User is not logged in
**Checklist**
- [x] Bug Report - The issue you are reporting might be there already. Please search on the [issues tracker](https://github.com/openfoodfacts/openfoodfacts-androidapp/issues) before creating or reporting a issue.
- [x] Before reporting, please ensure (if you can) that you have tested it with the latest (dev) version of the software.
- [x] Don't forget to tag your issue, it makes it easier to discover.
**Summary:** 
The app shows a popup of already logged in when the user is actually not logged in.

**Steps to reproduce:** 
* reinstall the latest build 
* open Sign in

**Expected behavior:** 
directly go to Sign in if the user is not signed in.

7.0 

**Screenshots:**

danzzen:
## Description

issue:   #996 Sign in option even after signing in.
Now logout option will appear until user don't logout.


app/src/main/java/openfoodfacts/github/scrachx/openfood/views/LoginActivity.java
```diff
@@ -92,22 +93,19 @@ protected void onCreate(Bundle savedInstanceState) {
 
     @OnClick(R.id.buttonSave)
     protected void attemptLogin() {
-        String login = loginView.getText().toString();
+        String login = loginView.getText().toString().trim();
```
Karljoones: This can not be trimmed, as discussed in the previous PR, whitespace is a valid character.

danzzen: Ok i will remove that part

app/src/main/java/openfoodfacts/github/scrachx/openfood/views/LoginActivity.java
```diff
@@ -147,7 +145,7 @@ public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response)
                 } else {
                     // store the user session id (user_session and user_id)
                     for (HttpCookie httpCookie : HttpCookie.parse(response.headers().get("set-cookie"))) {
-                        if (httpCookie.getDomain().equals(".openfoodfacts.org") && httpCookie.getPath().equals("/")) {
+                        if (httpCookie.getDomain().equals(".openbeautyfacts.org") && httpCookie.getPath().equals("/")) {
```
policorn: shouldn't this domain string be based on the flavor instead?

danzzen: But the Domain returned after login is this .openbeautyfacts.org

danzzen: @policorn sry i didn't get what you said. I printed the log and find that Domain name is .openbeautyfacts.org not the .openfoodfacts.org so comparing httpCookie.getDomain() with .openbeautyfacts.org will set the condition true and we will store the user and session and it will resolve the issue #996

Karljoones: I think what he is trying to say is that depending on the current flavour of the app, the domain should change.

If it is the OFF flavor, use `.openfoodfacts.org`
If it is the OBF flavor use `.openbeautfyfacts.org`

teolemon: .openpetfoodfacts.org for OPFF

policorn: Thanks @Karljoones, that's right; sorry @danzzen I wasn't more clear

danzzen: No problem @policorn

policorn:
Thanks @danzzen, @Karljoones! Merged

Issue #1042, opened 2018-03-02
Ensure Nutriscore in Search and History are disabled for OBF and OPFF


teolemon:
Ensure Nutriscore in Search and History are disabled for OBF and OPFF.
The Nutriscore is for OBF only, and we thus shouldn't display for OBF and OPFF

Shubham2320:
Okay I will work on this.

PR #1051, opened 2018-03-02
Ensured Nutriscore in Search and History are disabled for OBF and OPFF

Shubham2320:
## Description

Just added an if condition which checks the flavour before setting view of nutriscore in SumaryFragment
and as I've seen in History there was no view at present in which we're showing nutriscore, please correct me if I'm wrong.

Please suggest if any more changes are required :).

#1042


policorn:
Thanks @Shubham2320! Merged

PR #1062, opened 2018-03-03
Fixed #160 : Add a warning when scanning a book

### LINKED ISSUE #160: Add a warning when scanning a book
Books have a barcode that begins with 978 or 979
Warn that books are not currently supported by Open Food Facts (but don't prevent from adding).
https://en.wikipedia.org/wiki/Bookland

PrajwalM2212:
Description
Added a warning when scanning a book

Related issues and discussion
#160 

Screen Shot
screenshot_20180303-154527


teolemon:
We could have short barcodes beginning that way, so we'd need to check it's an EAN-13, but as it's a mere warning, that's not very important, I guess.

teolemon:
thanks @PrajwalM2212  :)

PR #1128, opened 2018-03-05
closes #445 Offer to type barcode if no barcode is detected after 30s

### LINKED ISSUE #445: Offer to type barcode if no barcode is detected after 30s
- Offer to type barcode if no barcode is detected after 30s in the Scan View. 
- Also add an option to dismiss forever
- If this option is chosen, add a final info that you can always search by barcode.

Shubham2320:
## Description

Implemented an Overlay on the screen of BarCodeScannerFragment which allows to search product using barcode if in 15 seconds no barcode is detected.

Barcode Scanner works also when the overlay appears, Overlay does not affect the functionality of Scanner.

closes #445
 
 ## Screen-shot:

Please suggest any Ux changes if needed :).


teolemon:
- Making the card side to side. 
- "Trouble scanning ? You can type the barcode here."

Shubham2320:
# ##Screenshot:
barcode final

@teolemon thanx for suggestions, I've made the changes please review.

Karljoones:
Code looks good. However, not really a huge fan of the dialog. How about something like this? I can share the code behind if it you need it

Shubham2320:
@Karljoones okay, I got this. I think I would need to use a library for it.

Karljoones:
@Shubham2320 you won't need a library, this is already built into android, check out [BottomSheetDialogFragments](https://developer.android.com/reference/android/support/design/widget/BottomSheetDialogFragment.html) :)

Shubham2320:
@Karljoones thanks, its really cool :).

Shubham2320:
@Karljoones 😄 changes are made now BottomSheetDialog is used instead of Dialog.

**Screenshot:**

sheetbottom

Karljoones:
Thank you! However, I don't think I am too much a fan of the buttons, what would it look like if there was only a search button? Maybe I am crazy though, @teolemon can I get your opinion on this?

teolemon:
I think you need to be able to dismiss it in some way. 

Cancel is intuitive, it could be swiped down, or a little grey cross on top of the sheet.

Karljoones:
@teolemon well the reason I asked that a bottom sheet fragment is used is that you can swipe down on it to cancel it, or touch outside of the fragment to dismiss.

app/src/main/res/layout/dialog_for_barcode.xml
```diff
@@ -0,0 +1,60 @@
...
+
+    <EditText
+        android:id="@+id/barcode"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:padding="@dimen/padding_short"
+       android:layout_margin="@dimen/spacing_tiny"
+        android:background="@color/white"
+        android:hint="@string/hintBarcode"
+        android:inputType="textEmailAddress" />
```
Karljoones: Can you change this to be "number". We want to make sure that the user can not put in any characters that may cause the app to crash. You might want to think about checking the output of this too, to ensure that people do not find a way of pasting some unsupported characters in here

Shubham2320:
@Karljoones @teolemon I've made some changes now the sheet is draggable also the user can drag sheet to enter barcode directly without waiting for 15 seconds.
one
expanded

Karljoones:
I really like this new design! However, it may be better to word it like

> Having trouble scanning?
> Drag up to type the barcode

Once that is changed, I think this is finally good to merge, thanks for all the work @Shubham2320!

Shubham2320:
finalbarcode
😄 finally its going to be merged, thanx @Karljoones for all the suggestions.

Karljoones:
Final final thing before I ask for one final check from another maintainer, what happens when the user swipes all the way down? Does this "hide" the fragment? I think that is what teolomon was referring to

Shubham2320:
yes, the user can fully dismiss the whole sheet by swiping it down.

Karljoones:
Fantastic! @teolemon final opinion please?

teolemon:
Perfect, merging :) 
thanks @Shubham2320 @Karljoones

Issue #1134, opened 2018-03-06
Not able to add Allergen


ManavPoddar:
**Summary:** 

Not able to add an Allergen.
![videotogif_2018 03 06_21 36 09](https://user-images.githubusercontent.com/18038492/37043622-1ae22258-2187-11e8-8e91-7dfd7fb2239b.gif)

**Screenshots:**

ManavPoddar:
I would love to work on it

ManavPoddar:
@teolemon  The PR #1145 fixes this issue. But for some languages like Hindi, there is no allergen for it in the allergen list. Shall I add a toast or dialog asking the user -> No allergens found, Please change the language ?

Shubham2320:
I think this issue would be resolved with the merging of #1115.

ManavPoddar:
Hey @Shubham2320 , #1115 resolves many issues, can you exactly tell which commit is solving the problem. Also the problem was because of error in the strings-hi.xml file as is discussed in issue #1144. Which is solved in PR #1145. So I think it can be merged

ManavPoddar:
So I think the issue can be closed after merging PR #1145

PR #1135, opened 2018-03-06
Fixed #1102: Languages should only be translated into their own language

l-o-b-s-t-e-r:
@Karljoones 
Implemented displaying of languages in the native languages.

Result: http://prntscr.com/inod56
Issue: #1102

PR #1145, opened 2018-03-07
Closes #1144 The labels of some languages in hindi are not correct

### LINKED ISSUE #1144: The labels of some languages in hindi are not correct
**Summary:** 

For some languages like English, French etc. These labels in hindi are set to the hindi names of these languages. Whereas it should be "en" for english, "fr" for french etc.

**Expected behavior:** 

Due to this, the label after changing the laguage to english back from hindi is set to अंग्रेज़ी. This causes problems in other parts of the app like the Add Alergens

**ScreenShots**
screenshot_20180307-111932

ManavPoddar:
## Description

Now the labels are correct. Also the allergen alert is working fine now

#1134 This is fixed Now

### DIFF
--- a/app/src/main/res/values-hi/strings.xml
+++ b/app/src/main/res/values-hi/strings.xml
@@ -29,13 +29,13 @@
   </string-array>
   <!-- Text units -->
   <string-array name="lang_array">
-    <item>अंग्रेज़ी</item>
-    <item>फ़्रेंच</item>
-    <item>स्पेनिश</item>
-    <item>इतालवी</item>
-    <item>जर्मन</item>
-    <item>जापानी</item>
-    <item>रूसी</item>
+    <item>en</item>
+    <item>fr</item>
+    <item>es</item>
+    <item>it</item>
+    <item>de</item>
+    <item>ja</item>
+    <item>ru</item>
     <item>pt</item>
     <item>sv</item>
     <item>my</item>


Karljoones:
This was fixed in PR #1135, no? Please make sure to check out the latest code to see if this was resolved.

ManavPoddar:
No @Karljoones , It is not fixed in PR #1135 , I have also checked with the latest code

Karljoones:
Just making sure! Wasn't 100% myself :)

teolemon:
Also for translations, you can correct them directly on Crowdin

l-o-b-s-t-e-r:
@Karljoones 
Yes, it wasn't fixed because I didn't even know that somebody could translate the international standard (country codes). So I just fixed translation of full county names.

l-o-b-s-t-e-r:
@Karljoones 
@teolemon 
This PR is good, but this is not enough. It fixes the only 1 language...but there is no guarantees that somebody won't translate country codes again in the future.

teolemon:
@l-o-b-s-t-e-r indeed. I'm sure we fixed it at some point, and not sure why it reappeared.

l-o-b-s-t-e-r:
@teolemon 
Because the country codes can be translatable. 
They are available for each language: https://crowdin.com/translate/openfoodfacts/83/en-uk, so every one can translate this.

We need to make them untranslatable directly in the android app (something like this: <string-array name="lang_array" translatable="false">)

PR #1178, opened 2018-03-09
Fixed #1134: Added option to change the language. 

### LINKED ISSUE #1134: Not able to add Allergen

ManavPoddar:
## Description

Since there were no allergens for some languages, so Added a new dialog which asks the user whether he wants to change the language.

#1134 
 
 ## Screen-shots, if any
 
videotogif_2018 03 09_11 47 46


ManavPoddar:
@teolemon  please take a look at this PR :)

Karljoones:
I'm not sure that this is the best way of handling this. I have looked through the allergens list that has been added to the app recently and can confirm that there are no Hindi translations included in the list, as there were no translations included in the source where the list was gotten from. There definitely needs to be a solution to this, but I just do not think that this is the best way to do it

ManavPoddar:
Hey @Karljoones, since there are no translations available, the app asks the users to download data for the translations. But again, after the download completes, the user is again  asked to download data. So I was thinking that we could notify the user about the unavailability of the allergens in his language. Shall I show a toast then instead of the dialog ? What do you suggest ?

Karljoones:
@ManavPoddar I think that would be best. Default to English translations for now and let them know that there are no translations available

ManavPoddar:
Opening a new PR, closing this one :)

PR #1215, opened 2018-03-12
fixes#1212 additives made in list

shivanshsoni:
Made additives in list.
screenshot_2018-03-12-07-31-20-904_openfoodfacts github scrachx openbeauty


Karljoones:
Merged, thanks!

PR #1218, opened 2018-03-12
fixes #1080 update method of DrawerItem's badge for OfflineEditFragment

### LINKED ISSUE #1080: Badge displays 1 while Offline screen is empty
image

Shubham2320:
## Description

Added update mechanism to update badge while uploading products online.

fixes #1080


Karljoones:
So on having a second glance over the code, I realised, this is only updated when the user opens the fragment. Is this correct? This badge should be updated even if the user does not open the fragment, this is to _draw_ the user into the fragment to do what they need to do there.

Shubham2320:
@Karljoones I think we only need to update the badge when we send products online  or delete any saved product otherwise there would be no change in the count of savedItems and we only do these things in OfflineEditFragment (please correct me here if I am wrong) so I've implemented this method by considering these senarios only. 
Do suggest me changes if necessary.

Karljoones:
@Shubham2320 this should be checked when the app is opened, updating this only when the user goes into the offline products tab could lead to the user not being notified. The user may not go into the offline products every time they open the app!

Shubham2320:
@Karljoones this also updates on MainActivity so everyTime when the user opens the app it gets updated.

Karljoones:
OK great in that case :)

Shubham2320:
Thanks.

teolemon:
@Karljoones Good to merge ?

Karljoones:
Merged, thanks @Shubham2320

PR #1161, opened 2018-03-08
fixes #1152 Made CommonActivity to Show list of products and made additive clickable.

Shubham2320:
## Description

- Removed SearchProductResultsFragment as we now have a common Activity to show the list of products.

- Changed the name of BrandActivity to ProductBrowsingListActivity which would be used to show the list of products.

- Made additives clickable.

## Related issues and discussion
closes #1152 
closes #1164
 
 ## Screen-shots, if any
 
additive

additivelist

PR #1226, opened 2018-03-13
Closes #1225, Closes #1234: Bug in search.

### LINKED ISSUE #1225: Bug in search.
**Summary:** 

On refreshing the search list by swiping the layout, multiple recycler view gets Stacked, the number of stacked recyclerview would be equal to the number of times layout is swiped, and we need to press back multiple times to get back to search results.
Also `onLoad()` method appends countOfProducts every time when it gets the `true` response which is should not happen.

**Steps to reproduce:** 

Use search------>Swipe layout to refresh multiple times---->Click on any item.

Now to get back to search results multiple times back need to be pressed.

**Screenshots:** 

**Product count is appended multiple times: **

screenshot_2018-03-13-06-18-37-438_openfoodfacts github scrachx openfood

### LINKED ISSUE #1234: Search results page behaving erratically
**Summary:** 

Search results page behaving erratically, loading or not loading

**Steps to reproduce:** 

Click on a facet

**Expected behavior:** 

Always display offline msg if offline
Always display results if online

Android Oreo - Google Pixel

**Screenshots:** 
image
image

Shubham2320:
## Description

Now instead of appending count of Products, setText is used.
created a Boolean which ensures that recyclerview does not get stacked.
Also added break statement in switch cases.
closes #1225
closes #1234


app/src/main/java/openfoodfacts/github/scrachx/openfood/views/ProductBrowsingListActivity.java
```diff
@@ -222,7 +224,7 @@ private void loadData(boolean isResponseOk, Search response) {
         if (isResponseOk) {
             mCountProducts = Integer.parseInt(response.getCount());
             if (pageAddress == 1) {
-                countProductsView.append(" " + NumberFormat.getInstance(getResources().getConfiguration().locale).format(Long.parseLong(response.getCount()
+                countProductsView.setText(getResources().getString(R.string.number_of_products)+ " " + NumberFormat.getInstance(getResources().getConfiguration().locale).format(Long.parseLong(response.getCount()
```
Karljoones: This is the only piece of code that needs to be changed to resolve this. Please ensure that the code is tested throughout the app lifecycle to ensure that the use of the `setUp` variable does not prevent this from handing `onResume()` and the rest of the app lifecycle

app/src/main/java/openfoodfacts/github/scrachx/openfood/views/ProductBrowsingListActivity.java
```diff
@@ -258,67 +260,73 @@ private void setUpRecyclerView() {
         offlineCloudLayout.setVisibility(View.INVISIBLE);
 
         productsRecyclerView.setVisibility(View.VISIBLE);
-        productsRecyclerView.setHasFixedSize(true);
-        LinearLayoutManager mLayoutManager = new LinearLayoutManager(ProductBrowsingListActivity.this, LinearLayoutManager.VERTICAL, false);
-        productsRecyclerView.setLayoutManager(mLayoutManager);
 
-        ProductsRecyclerViewAdapter adapter = new ProductsRecyclerViewAdapter(mProducts);
-        productsRecyclerView.setAdapter(adapter);
+        if (!setupDone) {
```
Karljoones: Please ensure that this will not prevent the list from being used throughout the app lifecycle

Shubham2320:
@Karljoones I've tested the app and it is working fine. Also, I've used boolean `setupDone` variable to ensure recycler view doesn't get stacked so that when we do swipe refresh multiple times and then click on any product, we can come back to search result list by pressing back button single time.

Shubham2320:
Travis-ci test is getting failed due to and the message is: 
The job exceeded the maximum time limit for jobs and has been terminated.
What should I do in this case?

policorn:
Illustrating product search issue:

bug

policorn:
PR causing/related:  #1161

Shubham2320:
Thanks, I will fix this.

Shubham2320:
@policorn I've tried running the app on emulator Nexus6 API22 and I've found that after typing query and pressing enter from the keyboard of laptop I get the exact same case which you told, but after typing query and pressing search from the keyboard of emulator the app worked fine.

I think that it is just happening on the emulator and if we just use emulator's screen to run the app without using laptop's keyboard then also the app will work fine.

Please provide your thoughts. 😄

Karljoones:
@Shubham2320 is there any chance you would be able to test this on a physical device? I know that this is not always the possible

Karljoones:
Fixed conflicts, going to merge once build is complete

Shubham2320:
@Karljoones finally the travis-ci build is passed please merge this one.

Karljoones:
Merged, thanks @Shubham2320

PR #1237, opened 2018-03-13
Fixed #439 : Disable the dropdown

### LINKED ISSUE #439: Disable the drowdown
https://github.com/mikepenz/MaterialDrawer/blob/develop/FAQ/accountheader_single_profile_without_dropdown.md

PrajwalM2212:
## Description

The dropdown has been disabled

#439 
 
## Screen Shots
d

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/MainActivity.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/MainActivity.java
@@ -142,6 +142,7 @@ public class MainActivity extends BaseActivity implements CustomTabActivityHelpe
                 .withTranslucentStatusBar(true)
                 .withHeaderBackground(R.drawable.header)
                 .addProfiles(profile)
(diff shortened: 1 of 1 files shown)


teolemon:
@Karljoones afterthoughs ? What's inside currently bugs for me. And Multiaccount is not very useful. I think we'll handle multiple people for the allergen detector, but directly in allergens.

Karljoones:
@teolemon I agree with this, it should be removed if it is not used. Plus the manage account button does not even work in there. Though if you did want to keep it, we could make sure that the manage account button works and move the logout button into the space that is there rather than having it as a menu item in the main menu. Up to you, at the moment though it is completely useless

teolemon:
I'm not sure how signing would work if we moved all user management to the top

Karljoones:
@teolemon could be tricky, let's merge this PR for now, that is something that can be added later

teolemon:
Merged. Thanks @PrajwalM2212 @Karljoones :)

PR #1250, opened 2018-03-15
Fixed #864 : Remove "no data" Toast from Offline edit fragment

### LINKED ISSUE #864: Remove "no data" toast from offline edit fragment
**Avoid reporting a duplicate**
- [x] Bug Report - The issue you are reporting might be there already. Please search on the [issue tracker](https://github.com/openfoodfacts/openfoodfacts-androidapp/issues) before creating reporting a issue.
   - Also before reporting please ensure that you have tested it with the latest (dev) version of the software.

**Summary:** 

When opening the offline edit fragment in the app, if there is no entries there, then there is a toast displayed with "no data". This should be removed in replace of a view being added into the center of the page. This view should tell the user that there are no offline products.

**Steps to reproduce:** 

Open the "offline edit" fragment.

**Expected behavior:** 

The app should inform the user that there are no products to display. However, there should be a view there instead of a toast, this view should always be visible when the user is in the fragment and there is no information to display.

**Observed behavior:** 

A toast is displayed.

OnePlus 5T Android 8

shushilkumarr:
## Description

- Added an ImageView and a TextView in the OfflineEditFragment layout
-  Added Code to toggle the visibility of the elements in the layout based whether the no. of offline products are "zero" or not
- Added code for "first offline fragment" that changes the text and image in the layout based on "firstUpload" boolean value stored in the shared prefference "usage"

#864 Remove "no data" Toast from Offline edit fragment
 
 ## Screen-shots, if any
 
alldone
firstoffline

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -8,6 +8,7 @@
     <string name="text_barcode">Type the barcode :</string>
     <string name="message_in_activity_save_product_offline">Only one photo is compulsory to send a new product.</string>
     <string name="no_offline_data">All your contributions have been sent.</string>
+    <string name="first_offline">Your offline contributions will appear here.</string>
 
 
     <!-- Nav Drawer Product Menu Items -->
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/OfflineEditFragment.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/OfflineEditFragment.java
@@ -116,6 +116,9 @@ public class OfflineEditFragment extends BaseFragment {
 
     @OnClick(R.id.buttonSendAll)
     protected void onSendAllProducts() {
+        SharedPreferences.Editor editor = getContext().getSharedPreferences("usage", 0).edit();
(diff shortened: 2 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -7,6 +7,8 @@
     <string name="action_find">Find</string>
     <string name="text_barcode">Type the barcode :</string>
     <string name="message_in_activity_save_product_offline">Only one photo is compulsory to send a new product.</string>
+    <string name="no_offline_data">All your contributions have been sent.</string>
```
Karljoones: Maybe this would sound better as
> You're all done!
Just something a little more upbeat and giving the user a sense of achievement

app/src/main/res/values/strings.xml
```diff
@@ -7,6 +7,8 @@
...
+    <string name="first_offline">Your offline contributions will appear here.</string>
```
Karljoones: And this to 
> Products added offline will appear here

Karljoones:
Maybe the icons should be more consistent with each other, check out [this](https://material.io/icons/#ic_cloud_done) and [this](https://material.io/icons/#ic_cloud_upload)

PR #1280, opened 2018-03-18
Fixed #1231, fixed #1261, fixed #1233: Gson + Performance issues. 

### LINKED ISSUE #1231: No translations + Duplicate additives
image

- The translation don't appear (fresh install, the translation file wasn't there yet, possibly)
- The additives appear duplicated.
- Erratic carriage return

### LINKED ISSUE #1233: Performance issues on the Product Page (at first load ?)
**Summary:** 

Performance issues on the Product Page (at first load ?)
No logs so far. Possibly had spotty network when it occured. Also, first load after fresh compile.

**Expected behavior:** 

Load the product fast, even on a budget phone

**Observed behavior:** 

Open Food Facts isn't responding, even on a premium phone.

Android Oreo, Google Pixel

**Screenshots:** 

image

### LINKED ISSUE #1261: Both Jackson and Gson library is used in app. Should be one of them.
At present both Jackson and Gson Library is used to handle JSON results whereas we can work with any one of them and remove the other which would help in reducing the APK's size.

l-o-b-s-t-e-r:
#1279 
 - fixed duplicated additives
 - added "synchronized" modifier for methods which save data. It improved performance a little bit.

#1261
 - removed Gson from the whole project. In cases of use Gson I switched to Jackson.

#1233
 - moved loading from the database to background thread. It probably improved performance.


teolemon:
@l-o-b-s-t-e-r I'm going to be the annoying voice here, but again, this could have been 3 or 4 different Pull Requests.
It would be awesome if you could break it down: 
- Jackson usage
- Additives issue
- Bkg thread
- Synchronizer method (ok to bundle with additives)

cc @ross-holloway94 @Karljoones

Issue #1285, opened 2018-03-19
OpenBeautyFacts crashes when Opening OpenFoodFacts via Intent


adamfils:
**Summary:** 
OpenBeautyFacts Crashes When Opening OpenFoodFacts Via Intent.

**Steps to reproduce:** 

Install O.B.F and O.F.F apps. Disable O.F.F in Phone Settings and open O.B.F.
Open the Navigation Drawer and click on "Open Food Facts".
The App Crashes.

**Add system logs:**

**Expected behavior:** 

I expected the App To Tell Me That O.F.F was disabled.
**Observed behavior:** 

The App Crashes Because startActivity has taken a null parameter.

**Device and Android version:** 

Tested on Galaxy S8(Android 7.0 stock rom) and LG Aristo(Android 7 Stock Rom)

**Screenshots:** 
![screenshot_2018-03-19-13-27-03](https://user-images.githubusercontent.com/22130093/37612020-5bd68838-2b7a-11e8-9399-b559988c97de.png)

I'll Like To Work On This

PR #1286, opened 2018-03-19
Fix App Crash in MainActivity.java

adamfils:
## Description

I put a null checking condition when firing an Intent in MainActivity.class to open the OpenFoodFacts app via The OpenBeautyApp.
Because If the OpenFoodFacts app is installed but disabled and the normal intent is fired, it crashes the app.

Describe the changes made and why they were made instead of how they were made.

#1285
 
 ## Screen-shots, if any
 
screenshot_2018-03-19-13-27-03

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/MainActivity.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/MainActivity.java
@@ -12,6 +12,7 @@ import android.graphics.Color;
 import android.net.Uri;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
+import android.provider.Settings;
 import android.support.annotation.NonNull;
 import android.support.customtabs.CustomTabsIntent;
 import android.support.v4.app.ActivityCompat;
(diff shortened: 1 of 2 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/views/MainActivity.java
```diff
@@ -257,7 +257,11 @@ public void onDrawerSlide(View drawerView, float slideOffset) {
                             if (otherOFAppInstalled) {
                                 Intent LaunchIntent = getPackageManager()
                                         .getLaunchIntentForPackage(BuildConfig.OFOTHERLINKAPP);
-                                startActivity(LaunchIntent);
+                                if (LaunchIntent != null) {
+                                    startActivity(LaunchIntent);
+                                }else{
+                                    Toast.makeText(this, "App Disabled", Toast.LENGTH_SHORT).show();
```
Karljoones: This needs to be a string resource
> Application is not installed

teolemon: Is that possible to propose reenabling the app ?

adamfils: Ok. I can write a function to alert the user that the app i currently disabled and suggest to take the user to the disabled app's setting to re-enable it.
What do you think @teolemon ?

PR #1291, opened 2018-03-20
Fixed #1285: App Crash in MainActivity.java 

### LINKED ISSUE #1285: OpenBeautyFacts crashes when Opening OpenFoodFacts via Intent

adamfils:
## Description

I put a null checking condition when firing an Intent in MainActivity.class to open the OpenFoodFacts app via The OpenBeautyApp, updated string resource file for toast displayed and also made an intent to open up the app settings of the Disabled App.
Because If the OpenFoodFacts app is installed but disabled and the normal intent is fired, it crashes the app.

#1285 
 
 ## Screen-shots, if any
 
screenshot_2018-03-19-13-27-03

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -486,6 +486,6 @@
     <string name="category_string">Category</string>
     <string name="connection">Connected</string>
     <string name="contributor_string">Contributor</string>
-    <string name="app_disabled_text">Application Is Disabled</string>
+    <string name="app_disabled_text">Application is disabled</string>
 
 </resources>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -486,6 +486,6 @@
     <string name="category_string">Category</string>
     <string name="connection">Connected</string>
     <string name="contributor_string">Contributor</string>
-    <string name="app_disabled_text">Application Is Not Installed</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -486,5 +486,6 @@
     <string name="category_string">Category</string>
     <string name="connection">Connected</string>
     <string name="contributor_string">Contributor</string>
+    <string name="app_disabled_text">Application Is Disabled</string>
```
Karljoones: Please remove capitalisation, "Application" should only start with a capital

teolemon: "The application is disabled."

PR #1302, opened 2018-03-21
Fixed #1261: Both Jackson and Gson library is used in app.

### LINKED ISSUE #1261: Both Jackson and Gson library is used in app. Should be one of them.

l-o-b-s-t-e-r:
#1261
Removed Gson from the whole project. In cases of use Gson I switched to Jackson.


Karljoones:
Merged thanks @l-o-b-s-t-e-r

PR #1324, opened 2018-03-23
Fixes #422 Add an option to show barcode

### LINKED ISSUE #422: Add an option to show the barcode in product view
Add an option to show the barcode as an image in product view (for people who need for some reason to scan the barcode of a product)

ManavPoddar:
## Description

Now we can generate a QR code for users to scan it if they need to !!

#422 
 
 ## Screen-shots, if any
 
videotogif_2018 03 23_23 35 43


app/build.gradle
```diff
@@ -158,9 +158,13 @@ dependencies {
     //CSV Serialization/Deserialization
     implementation "com.opencsv:opencsv:$rootProject.openCsvVersion"
 
+
     //Database
     implementation "org.greenrobot:greendao:$rootProject.greenDaoVersion"
 
+    //image popup
+    implementation 'com.github.chathuralakmal:AndroidImagePopup:1.2.1'
```
Karljoones: There seems to be a lot of extra libraries added into this. It might just be best to use the image viewer that is already build into the app (the same as is used to display the product images). Can you try this rather than using this library?

teolemon:
Also, I was expecting it to show the actual barcode, rather than the QR-Code

ManavPoddar:
Hey @teolemon  is there any library or api call i can use to generate the actual barcode ?

Karljoones:
@ManavPoddar the barcode can be gotten from the scanning of the product, or the `code` field from the JSON

ManavPoddar:
Yeah @Karljoones, from what I can get is that EAN_13 barcode format is being used. So shall I show a 1D barcode instead of the QR code and just simply display it rather than adding a button for it ? What do you suggest ?

Karljoones:
I get what you mean. I don't think that is the kind of feature that we need, a simple pop up of the barcode (numbers) or displaying it in the app would suffice. Not an actual image of the barcode itself.

ManavPoddar:
Hi @Karljoones, There's already a barcode(numbers) in the product view of the application. So can this issue be closed now?
screenshot_20180330-223545

Karljoones:
Sure @ManavPoddar, thank you for the work you done for this!

Issue #1327, opened 2018-03-26
E/SQLiteLog: (1) no such column: T.WIKI_DATA_ID


teolemon:
03-26 12:20:43.265 8077-8077/openfoodfacts.github.scrachx.openfood E/SQLiteLog: (1) no such column: T.WIKI_DATA_ID
03-26 12:20:43.266 8077-8077/openfoodfacts.github.scrachx.openfood D/AndroidRuntime: Shutting down VM
03-26 12:20:43.269 8077-8077/openfoodfacts.github.scrachx.openfood E/AndroidRuntime: FATAL EXCEPTION: main
                                                                     Process: openfoodfacts.github.scrachx.openfood, PID: 8077
                                                                     android.database.sqlite.SQLiteException: no such column: T.WIKI_DATA_ID (code 1): , while compiling: SELECT T."_id",T."CATEGORY_TAG",T."LANGUAGE_CODE",T."NAME",T."WIKI_DATA_ID",T."IS_WIKI_DATA_ID_PRESENT" FROM "CATEGORY_NAME" T  WHERE T."CATEGORY_TAG"=? AND T."LANGUAGE_CODE"=?
at android.database.sqlite.SQLiteConnection.nativePrepareStatement(Native Method)
at android.database.sqlite.SQLiteConnection.acquirePreparedStatement(SQLiteConnection.java:890)
at android.database.sqlite.SQLiteConnection.prepare(SQLiteConnection.java:501)
at android.database.sqlite.SQLiteSession.prepare(SQLiteSession.java:588)
at android.database.sqlite.SQLiteProgram.<init>(SQLiteProgram.java:58)
at android.database.sqlite.SQLiteQuery.<init>(SQLiteQuery.java:37)
at android.database.sqlite.SQLiteDirectCursorDriver.query(SQLiteDirectCursorDriver.java:46)
at android.database.sqlite.SQLiteDatabase.rawQueryWithFactory(SQLiteDatabase.java:1392)
at android.database.sqlite.SQLiteDatabase.rawQuery(SQLiteDatabase.java:1331)
at org.greenrobot.greendao.database.StandardDatabase.rawQuery(StandardDatabase.java:32)
at org.greenrobot.greendao.query.Query.unique(Query.java:129)
at org.greenrobot.greendao.query.QueryBuilder.unique(QueryBuilder.java:479)
[...]

PR #1338, opened 2018-03-26
Solved bug appeared in schema in implementation of wikidata

Shubham2320:
## Description

Schema version is changed to 7

Issue #1327 
modified PR #1332

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -106,7 +106,7 @@ buildscript {
 apply plugin: 'org.greenrobot.greendao'
 
 greendao {
-    schemaVersion 6
+    schemaVersion 7
 }
 android {
     // Other settings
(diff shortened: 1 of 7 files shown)


Shubham2320:
@teolemon @Karljoones @l-o-b-s-t-e-r  please review it.

teolemon:
poke @l-o-b-s-t-e-r ;-)

l-o-b-s-t-e-r:
Yeah, it works.
One more thing which should be done while upgrading is setting shared property "last_refresh_date" -> 0, to make sure that application will load data again and fill WikiData. Just open the SplashPresenter and look at "isNeedToRefresh()" method to understand how it's used.

To tell the truth after deep investigation of code related to WikiData issue, I have a few suggestions about implementation. Let me know if somebody is interested in my opinion.

Shubham2320:
It gets updated int the splashPresenter itself in the same way in which taxonomies gets updated so there is no need to do that.

l-o-b-s-t-e-r:
If you don't do it, Application will refresh data only after 24hours from the last update date, but we need to update it right after installing the new version.

Shubham2320:
Ohh I see, so instead of adding a new variable, I would like to change the string of `last_refresh_date` in shared preference so.

l-o-b-s-t-e-r:
But you didn't set it = 0 while upgrading :D

Shubham2320:
could you explain why it is required?

l-o-b-s-t-e-r:
Okay, we figured it out. Everything is ok, application will update data immediately after installing the new version :)

PR #1402, opened 2018-04-05
GreenDao upgrade

l-o-b-s-t-e-r:
Implemented faster way to save taxonomies translations.
Previously, to save each record, GreenDao used a new transaction, it had awful effect on performance.
At the moment, GreenDao saves records using only 1 transaction.
Look at "Performance" header: http://greenrobot.org/greendao/documentation/technical-faq/


ross-holloway94:
@teolemon I'll leave the merge decision to yourself.

PR #1419, opened 2018-04-07
Fixed  #1409 : The edit icon on the new bottom nav has been replaced by a search icon.

### LINKED ISSUE #1409: The edit icon on the new bottom par has been replaced by a search icon.
The edit icon on the new bottom par has been replaced by a search icon.
The crayon is much more explicit.

PrajwalM2212:
## Description
1. I have added the edit icon as desired
2. Since the bottom nav was feeling a bit empty , i have added search and history

#1409 
 
 ## Screen-shots, if any
 
screenshot_1523043276


teolemon:
Nice touch, it was indeed empty. Can you add text for the 3 other icons ?

PrajwalM2212:
@teolemon  The text is displayed for only selected item. That is the default behaviour . Google does not allow us to display text for all .
To display text for all icons we have to forcefully implement it and change pro-guard configuration as it will use reflection .

Do you like the previous design or this ?
image

PrajwalM2212:
@teolemon  Do you need gray or white background for bottom navigation ?

huzaifaiftikhar:
@PrajwalM2212 It seems that by default if we have more than 3 items on the bottom menu, the title is hidden due to the shift animation of the title text. However, you can easily change that and make it work according to our need. Refer [this](https://stackoverflow.com/questions/41718633/bottomnavigationview-with-more-than-3-items-tab-title-is-hiding/41718722?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa).

PrajwalM2212:
@huzaifaiftikhar  The method in your link is also the same answer but does not warn about using reflection . Since we are using reflection , it will cause obfuscation problems .

teolemon:
Can you rebase on latest trunk ? It should fix the build.

teolemon:
@PrajwalM2212  Can you rebase this ?

app/src/main/res/values/strings.xml
```diff
@@ -134,7 +134,7 @@
 
     <!-- Content Description -->
     <string name="bookmark">Bookmark</string>
-    <string name="product">Product</string>
+    <string name="product">Edit</string>
```
Karljoones: Can the actual string id be changed to? This may cause confusion in the future

app/src/main/res/values/strings.xml
```diff
@@ -551,6 +551,8 @@
     <string name="format_error">Wrong Barcode/QR format</string>
     <string name="code_detected">We have detected the barcode</string>
     <string name="do_you_want_to">Do you want to add photos to this barcode?</string>
+    <string name="bottomNav_history">History</string>
+    <string name="search_bottomNav">Search</string>
```
Karljoones: Can these two strings be standardised? They both flip where `bottomNav` is placed in the ID.

Karljoones:
@PrajwalM2212 any chance you could have a quick look at this PR too and fix the conflicts so that I can merge it?

PrajwalM2212:
@Karljoones I will look into it

Karljoones:
@PrajwalM2212 thank you!

PrajwalM2212:
@Karljoones  Done

PrajwalM2212:
@Karljoones  I have made a minor change . The travis should build now.

huzaifaiftikhar:
The spacing between the buttons doesn't seem equal due to an empty item added. I am sure this item is added so that the rightmost icon doesn't hide behind the scan FAB. Isn't there another method to overcome this rather than adding an empty item in the code which doesn't seem right!

teolemon:
There's a more fundamental problem with the bottom bar to be honest:
https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/1517

But this improves the current design.

Karljoones:
Bloop, merged, thanks @PrajwalM2212. We can look into the other problems by building on this

PR #1441, opened 2018-04-11
updated scannerfragment intent flags

### LINKED ISSUE #1408: After adding multiple products, you need to click multiple times on the back menu to go back to settings.
After adding multiple products, you night to click multiple times on the back menu to go back to settings.
You should only have to click once. 
High priority.

jaztriumph:
## Description

Added clear top flag for scanner fragment so that only the first one is stored and give us the required functionality.

closes #1408 
 
 ## Screen-shots, if any
 
previously we had to press multiple times to go to home after scanning many products, something like below

backstack-bugged

After adding flags

backstacksolved


Karljoones:
Merged, thanks @jaztriumph

PR #1445, opened 2018-04-14
removed confirmation to download

### LINKED ISSUE #1355: Auto download allergen file (don't ask for confirmation)
Currently, there is a confirmation step for the allergen file download. We should attempt to autodownload it, like we do for the taxonomies.

jaztriumph:
## Description

Removed the dialog for confirmation to download of allergens and then notified the change in data set.

closes #1355


teolemon:
Can you rebase on latest trunk ? It should fix the build.

Karljoones:
Merged, thanks @jaztriumph

PR #1508, opened 2018-04-25
More unit tests for the models.

gspitz01:
## Description

Some more unit tests for the models. I also fixed some of the formatting on the unit tests I did two weeks ago. I didn't touch any of the production code. I did add some TODO's, though, in the tests that suggest changes to the production code, which I will try to implement soon.

None. Doesn't touch production code.


gspitz01:
It looks like I screwed up on the fetching / comitting. Seems like there's an extra commit in there. I can fix that if needed.

Karljoones:
Merged, thanks @gspitz01

PR #1577, opened 2018-05-15
Closes #287 : Change wording of offline mode warning popup for Open Beauty Facts.

### LINKED ISSUE #287: Change wording of offline mode warning popup for Open Beauty Facts
image

Depends on #279

jaindiv26:
## Description
Changed wording of offline mode warning popup for Open Beauty Facts from "You can scan products when you\'re offline to add them to Open Food Facts later. All the products you scan will appear here, and you\'ll be able to send them using the button below." to "Scan the products even when you are offline so as to add them to Open Beauty Facts later. The scanned products will appear here and you will be able to send them using the button below. ".

Change wording of offline mode warning popup for Open Beauty Facts
Issue Number: #287 
 
 ## Screen-shots, if any
 
screenshot_1526409083


jaindiv26:
@huzaifaiftikhar a string resource name "text_offline_info_dialog" already exist. Just to be clear you want me to keep the original string intact without any changes and create a new string resource "XYZ" and put the changed wording and use them in the toast?

huzaifaiftikhar:
@jaindiv26 The string resource named "text_offline_info_dialog" should be moved from ...\src\main\res\values\strings.xml to ...\src\obf\res\values\strings.xml. Similarly it should also be moved to ...\src\off\res\values\strings.xml and so on for opf and opff too! 
In each of these 4 versions of the same string name you should put the value as required by the app. For obf a different value, for off a different one and so on.

PR #1698, opened 2018-06-12
Dirty Oreo service fix

teolemon:


huzaifaiftikhar:
We need to add this too
```
@Override
public void onCreate() {
    super.onCreate();
    startForeground(1,new Notification());
}
```

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/WifiUploadReceiver.java
```diff
@@ -40,13 +40,14 @@ public void onReceive(Context context, Intent intent) {
...
 
-
     }
-
+    @Override
+    public void onCreate() {
+        super.onCreate();
+        startForeground(1,new Notification());
+    }
     public static class WifiService extends Service {
```
huzaifaiftikhar: This needs to be added inside the WifiService class

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/WifiUploadReceiver.java
```diff
@@ -32,18 +32,29 @@ public void onReceive(Context context, Intent intent) {
...
-
+            //context.startService(new Intent(context, WifiService.class));
+            // Quick fix. TODO: Fix it the right way: https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/1583
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) { 
+                context.startForegroundService(new Intent(context, ServedService.class)); 
+            } else {
+                context.startService(new Intent(context, ServedService.class));
+            }
+            
         }
```
huzaifaiftikhar: `ServedService.class` should be replaced with `WifiService.class` as it was before.

PR #1774, opened 2018-07-14
Fixes #1634 and user can calculate nutrition facts of a product

### LINKED ISSUE #1634: Calculate all lines of the nutrition table with the Calorie calculator
Discussed with someone who is diabetic yesterday. She would love to have all the values of the nutrition table calculated for the portion in the calorie calculator, because she needs glucose level for the portion she's eating.

Rutvikmodi:
## Description

As @teolemon suggested in #1634 that user should be able to calculate every nutrition fact of nutrition present in the "Nutrition Facts", so menu item for all of them are created and only items which are visible in "Nutrition Facts" are displayed.


teolemon:
Can you add the screenshot you send me in your PR ? @Rutvikmodi

Rutvikmodi:
screenshot_2018-07-15-00-55-52-676_openfoodfacts github scrachx openfood

On clicking any menu item we will be able to see the same dialog box as of we find it in "calculate calorie" and we can calculate any nutrition fact.

app/src/main/res/values/strings.xml
```diff
@@ -313,31 +313,51 @@
     <string name="calculate_calories">Calculate Calories</string>
     <string name="txt_calories_result">%1$s kcal in %2$s %3$s</string>
     <string name="nutrition_carbohydrate">Carbohydrate</string>
+    <string name="calculate_carbohydrate">Calculate Carbohydrate</string>
+    <string name="txt_carbohydrate_result">%1$s %2$s in %3$s %4$s</string>
```
Karljoones: For this, you should have a general template. This means that you should only have one version of this string rather than the 20 or so that are there. Since they are all the same string.

teolemon:
How about:
"Calculate values for custom weight" and displaying all values for that weight ?
We have a quasi infinite number of ingredients.

app/src/main/res/values/strings.xml
```diff
@@ -312,54 +312,40 @@
     <string name="no_title">No Title</string>
 
     <!-- Nutrition info tab -->
-    <string name="calculate_calories">Calculate Calories</string>
+    <string name="check_facts">Check!!</string>
```
teolemon: We should avoid exclamation marks.

ross-holloway94:
@Karljoones what do you think?

Karljoones:
Haven't been able to run the code myself due to other commitments. From my understanding, there is a new activity that all the information is shown in, can we get a screenshot of this new screen, @Rutvikmodi ?

Rutvikmodi:
screenshot_2018-07-22-22-56-07-917_openfoodfacts github scrachx openfood

PR #1812, opened 2018-08-07
Closes #1778 : Implement Google smart lock for login

### LINKED ISSUE #1778: Suggestion: Implement Google smart lock for login
> If you've never installed an app that supports Smart Lock, you may not realize how seamless it is, even moreso than relying on Autofill. You open the app the first time and if you've ever logged into that service from Chrome or another Android phone with Smart Lock enabled, you get a little pop-up asking you if you'd like to login using those saved credentials. No tapping on the user/email field to fill them up, no waiting and unlocking whichever Autofill app you use. It's a simple tap and you're logged in.

As suggested [here](https://www.androidpolice.com/2018/03/05/google-smart-lock-passwords-underused-underrated-wish-android-developers-implemented/)

jaindiv26:
## Description
Added Google Smart Login for Password in LoginActivity.

#1778 : Suggestion: Implement Google smart lock for login 
 
 ## Screen-shots, if any
 
screenshot_1533575260
screenshot_1533625485
screenshot_1533625506

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -142,6 +142,7 @@ dependencies {
     implementation "com.android.support:support-annotations:$rootProject.supportLibraryVersion"
     implementation 'com.android.support.constraint:constraint-layout:1.1.2'
     implementation "com.android.support:multidex:$rootProject.multidexVersion"
+    implementation 'com.google.android.gms:play-services-auth:15.0.1'
 
     //DI
     annotationProcessor "com.google.dagger:dagger-compiler:$rootProject.daggerVersion"
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/LoginActivity.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/LoginActivity.java
@@ -3,6 +3,7 @@ package openfoodfacts.github.scrachx.openfood.views;
 import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
+import android.content.IntentSender;
 import android.content.SharedPreferences;
 import android.content.pm.ActivityInfo;
 import android.hardware.Sensor;
@@ -11,6 +12,7 @@ import android.net.Uri;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.support.customtabs.CustomTabsIntent;
 import android.support.design.widget.Snackbar;
 import android.support.v4.content.ContextCompat;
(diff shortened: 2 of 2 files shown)


teolemon:
What happens if the user doesn't have the Google Apps on his phone ?

app/build.gradle
```diff
@@ -142,6 +142,7 @@ dependencies {
     implementation "com.android.support:support-annotations:$rootProject.supportLibraryVersion"
     implementation 'com.android.support.constraint:constraint-layout:1.1.2'
     implementation "com.android.support:multidex:$rootProject.multidexVersion"
+    implementation 'com.google.android.gms:play-services-auth:15.0.1'
```
teolemon: Is it a non free dependency ? If so, we should only enable it with a flag.
--build-with-non-free-components

jaindiv26:
@teolemon I'll put a check on Google Apps availability, and I don't know much about a non-free dependency, can you please guide me.

teolemon:
https://stackoverflow.com/questions/15401748/how-to-detect-if-google-play-is-installed-not-market/19955415#19955415

jaindiv26:
@teolemon Any Updates!

app/build.gradle
teolemon: @jaindiv26 I'm concerned about all those imports. Can't they be made conditional @Karljoones @ross-holloway94 ?

app/src/main/java/openfoodfacts/github/scrachx/openfood/views/LoginActivity.java
```diff
@@ -24,6 +26,20 @@
...
+import com.google.android.gms.auth.api.credentials.CredentialsClient;
+import com.google.android.gms.auth.api.credentials.CredentialsOptions;
+import com.google.android.gms.common.ConnectionResult;
+import com.google.android.gms.common.GoogleApiAvailability;
+import com.google.android.gms.common.api.ApiException;
+import com.google.android.gms.common.api.CommonStatusCodes;
+import com.google.android.gms.common.api.GoogleApiClient;
+import com.google.android.gms.common.api.ResolvableApiException;
+import com.google.android.gms.tasks.OnCompleteListener;
+import com.google.android.gms.tasks.Task;
```
teolemon: @jaindiv26 I'm concerned about all those imports. Can't they be made conditional @Karljoones @ross-holloway94 ?

Karljoones: @teolemon Java does not support conditional imports, so that would not be possible.

teolemon: so not possible to bind them to a non free flavour ?

jaindiv26: @teolemon Yes, it will not be possible.

Karljoones:
I'm going to close this off for now. Sorry @jaindiv26, thank you so much for your work but I think this would break other flavours of the app that do not rely on Google Apps.

jaindiv26:
@Karljoones That's why I added GooglePlayServicesCheck so there won't be any break.

PR #1850, opened 2018-08-19
linting

teolemon:
Some refactoring after using the Linting feature of Android Studio


ross-holloway94:
Looks like you have a TODO in fragment_contributors? Otherwise all fine I think

PR #1867, opened 2018-09-07
Fixes #1864 Encoding error on the Product URL field search function

### LINKED ISSUE #1864: Encoding error on the Product URL field search function
Encoding error on the search function within the Product URL field in the edit mode

image

brandenfung:
## Description

The problem was caused by calling getText().toString() on a NachoTextView with chipified text, which would result in special characters that messed up the query. The proper way to get the chipified text is to call NachoTextView.getChipValues().

Fixes #1864

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AddProductOverviewFragment.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AddProductOverviewFragment.java
@@ -792,8 +792,9 @@ public class AddProductOverviewFragment extends BaseFragment {
     @OnClick(R.id.hint_link)
     void searchProductLink() {
         String url = "https://www.google.com/search?q=" + code;
-        if (!brand.getText().toString().isEmpty()) {
-            url = url + " " + brand.getText().toString();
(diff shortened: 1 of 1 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AddProductOverviewFragment.java
```diff
@@ -793,7 +793,8 @@ void toastEmbCodeHint() {
     void searchProductLink() {
         String url = "https://www.google.com/search?q=" + code;
         if (!brand.getText().toString().isEmpty()) {
```
huzaifaiftikhar: The condition could also be changed to `!brand.getChipValues().isEmpty()`

huzaifaiftikhar: This looks like a corner case however I don't think it would cause any issue so we can consider it as the intended behavior.

brandenfung: Okay, done. I've used `brand.getChipAndTokenValues()` instead so both chipped and non chipped brand names are in the query.

PR #1871, opened 2018-09-10
Fixes #1846 The allergen field behaves weirdly

### LINKED ISSUE #1846: The allergen field behaves weirdly
The allergen field behaves weirdly
screenshot_20180816-231247

brandenfung:
## Description

The weird behaviour of allergen field was caused by improper regex matching of allergen names, which caused issues with allergens with multi-part names and improper spacing. For example the matching result for "lécithine de soja, blé" was ["lécithine ", "de ", "soja", "blé" ].

I've modified the regex to take this into consideration and to use commas a delimiters.

Fixes #1846
 
 ## Screen-shots, if any
image


teolemon:
Looks good to me. Also a nice thing in the future would be to make it use the taxonomy/system we use for categories/labels.

Issue #1625, opened 2018-05-24
The first image is not displayed in the "Share to Open Food Facts" modal 


teolemon:
![image](https://user-images.githubusercontent.com/1689815/40485330-86be7dfe-5f5e-11e8-994a-979a7389a857.png)
The first image is not displayed in the "Share to Open Food Facts" modal. It didn't have a barcode.

teolemon:
Maybe if no barcode is detected within it.

brandenfung:
Added in #1873

Issue #1599, opened 2018-05-20
Share to Open Food Facts fails when the language is not English


teolemon:
Share to Open Food Facts fails when the language is not English

jaindiv26:
@teolemon In which language app fails and what build variant are you using obf, off, opf, opff ?

teolemon:
OFF, French fails for instance.

jaindiv26:
I've tried changing many many languages and share the OFF content to many platforms (Whatsapp, Facebook, SMS, copy to clipboard) but never faced this issue! Well, at this point I can't seem to reproduce this issue if I'm missing something so as to reproduce this issue do let me know.

teolemon:
It's inbound sharing (Share to, and not from). You can share photos from
your photo roll to Open Food Facts, and if they have a barcode, Open Food
Facts will try to add them to the right product.
----------------------------------------
Pierre Slamich - pierre@slamich.net - (+33) 6 02 13 14 57

jaindiv26:
@teolemon Do you also get the same result when the barcode is detected and when you click on the yes button, a toast pops up stating "Image will be uploaded when the network is available." but you are already connected to the network?

teolemon:
Yes. That's another bug, even in English.

jaindiv26:
@teolemon I'll try to solve this issue asap.

brandenfung:
@teolemon Can you describe how Share to Open Food Facts fails? Does the app crash?

teolemon:
So you need (phone/off ?) to be in German, French, (eg not English) and share one or several images, including one with a barcode and select "Share to Open Food Facts".
The expected result is that it will either 
- ask you to type manually the barcode (if no barcode was detected) (and the proceed to next steps)
- add the pictures to an existing product (if a barcode is detected and the product exists)
- add the pictures to a new product (if a barcode is detected and the product does not exists)

teolemon:
In languages other than english; it will fail silently.
this feature is very useful for Scan parties, when users take many pictures using their camera

brandenfung:
Does this #1873 fix this issue?

teolemon:
I'm still not able to upload products (will be uploaded when you get network), but at least the modal now appears everytime.

Issue #1520, opened 2018-04-27
Show shared photos, even if a barcode has been detected.


teolemon:
![image](https://user-images.githubusercontent.com/1689815/39354052-b8722f04-4a09-11e8-931c-e3f4f35228f0.png)
Show shared photos, even if a barcode has been detected, probably as a scrollable horizontal carousel ?

brandenfung:
Added in #1873

PR #1873, opened 2018-09-12
Fixes for Share to Open Food Facts

### LINKED ISSUE #1518: "Share to Open Food Facts" fails silently
- Share a photo with barcode to Open Food Facts
- The app probably doesn't recognize any barcode
- It stays on the homepage without any message to type manually the barcode (already exists within the code)
- The app was in Italian

brandenfung:
## Description
Added horizontal scrolling recycler view for viewing images shared to off. Photos are shown for both detected and undetected barcode. Fixed bug where sharing to open food facts would fail silently on cold launch of app (maybe fixes #1518 / #1599?)

#1625 
#1599
#1518
#1520 
 
 ## Screen-shots, if any
image

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/MainActivity.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/MainActivity.java
@@ -535,6 +535,7 @@ public class MainActivity extends BaseActivity implements CustomTabActivityHelpe
             apiClient.syncOldHistory();
         }
 
+        handleIntent(getIntent());
     }
 
     private void scan() {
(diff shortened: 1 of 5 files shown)


teolemon:
Just tested. Your fixes solve the not appearing part. 
- The network call to actually send the picture seems to fail, but that was a separate (already filed) issue. 
- Another (already filed) separate issue is the ability to pinch and zoom on a picture to be able to see an undetected barcode (pinch and zoom so that you can type it at the same time)

teolemon:
Thanks a lot @brandenfung :)
Merged.

PR #1885, opened 2018-09-16
fix rating

PrajwalM2212:
Fix rating issue


PrajwalM2212:
screenshot_1537086718

PR #1913, opened 2018-09-22
Fixes #1519, closes #1373: "Share to Open Food Facts" always says it will upload the picture when the network is available

### LINKED ISSUE #1373: Improvements to the share with Open Food Facts feature
- Always says it will upload the picture when the network is available (doesn't create any offline product, and the network is actually available in those cases)
- Doesn't work when selecting two or more pictures
- It should show the shared picture even when a barcode is detected
- It should let you zoom on the picture when no barcode is detected
- The Share menu should be named "Add to Open Food Facts" instead of just "Open Food Facts"

### LINKED ISSUE #1519: "Share to Open Food Facts" always says it will upload the picture when the network is available
Always says it will upload the picture when the network is available (doesn't create any offline product, and the network is actually available in those cases)

image

image

brandenfung:
## Description
Fixed improper retrieval of image path from a content Uri that was used to upload the image.

Fixes #1519 
Closes #1373

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -62,14 +62,16 @@
             <intent-filter>
                 <action android:name="android.intent.action.SEARCH" />
             </intent-filter>
-            <intent-filter>
+            <intent-filter
+                android:label="@string/intent_filter_share_label">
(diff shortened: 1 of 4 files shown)


teolemon:
Merged. Thanks @brandenfung and @Karljoones

PR #1983, opened 2018-10-23
Show additive overexposure table even if wikidata is missing

ripetrescu:
Signed-off-by: Rares Petrescu <petrescu.rares@gmail.com>

Change the default behaviour when clicking a moderate/high additive by showing the overexposure table even if wikidata is missing.

Before this change, if the wikidata was missing, a click on the additive name will automatically search for products containing the additive.
With this patch, if wikidata is missing, we hide the description and the wiki button from additive details popup, but we still display the overexposure table and the search button.

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/product/ingredients/IngredientsProductFragment.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/product/ingredients/IngredientsProductFragment.java
@@ -336,7 +336,9 @@ public class IngredientsProductFragment extends BaseFragment implements IIngredi
                             }
                         } else {
                             if (additive.hasOverexposureData()) {
-                                productActivity.showBottomScreen(null, additive);
+                                if (productActivity != null && !productActivity.isFinishing()) {
+                                    productActivity.showBottomScreen(null, additive);
+                                }
                             } else {
                                 ProductBrowsingListActivity.startActivity(getContext(), additive.getName(), SearchType.ADDITIVE);
                             }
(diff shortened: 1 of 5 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/views/product/ingredients/IngredientsProductFragment.java
```diff
@@ -70,569 +75,675 @@
...
-            }
-
-        } else {
-            // take a picture
-            if (ContextCompat.checkSelfPermission(getActivity(), CAMERA) != PERMISSION_GRANTED) {
-                ActivityCompat.requestPermissions(getActivity(), new String[]{CAMERA}, MY_PERMISSIONS_REQUEST_CAMERA);
-            } else {
-                EasyImage.openCamera(this, 0);
+public class IngredientsProductFragment extends BaseFragment implements IIngredientsProductPresenter.View
+{
```
brandenfung: why is the formatting changed in this file

ripetrescu: You are right. I was working on a different project that is using a different formatting and I forgot to switch it back before pushing the patch. So I reformatted again and updated the mr.
Thanks.

app/src/main/java/openfoodfacts/github/scrachx/openfood/views/product/ingredients/IngredientsProductFragment.java
```diff
@@ -316,63 +321,64 @@ public void refreshView(State state) {
...
-                                }
+                    apiClientForWikiData.doSomeThing(additive.getWikiDataId(), (value, result) -> {
+                        ProductActivity productActivity = (ProductActivity) getActivity();
+                        if (value) {
+                            if (productActivity != null && !productActivity.isFinishing()) {
+                                productActivity.showBottomScreen(result, additive);
+                            }
+                        } else {
+                            if (additive.hasOverexposureData()) {
+                                productActivity.showBottomScreen(null, additive);
```
brandenfung: null check & isFinishing check for productActivity

ripetrescu: Done. Pushed a new patch.

teolemon:
Merging, thanks @ripetrescu and thanks @brandenfung 👍

PR #1995, opened 2018-11-10
Removed unused string resources

### LINKED ISSUE #1991: Check whether those strings are still in use
```
    <string name="seconds_ago">seconds ago</string>
    <string name="minutes_ago">minutes ago</string>
```
Check whether those strings are still in use. If they are still in use, they should be using placeholders and probably plural/singular.
Also, if so change the name of the string to avoid a bug with persistant translations within crowdin

brandenfung:
## Description
Removed the unused string resources

closes #1991


teolemon:
Looks good to me.

teolemon:
Thanks, merged.

Issue #2007, opened 2018-11-17
Adding a tweakable call to action in home


teolemon:
Adding a tweakable call to action in home that would use a file on the Open Food Facts server, and that we could update on the fly.

![image](https://user-images.githubusercontent.com/1689815/48661748-66b5d300-ea77-11e8-8685-fa1dbe6aa669.png)

jaindiv26:
@teolemon Here's just a sample how it would look, Do we also need to update this textView if not please give a suggestion about what would be the proper text for this, and also I need the API that would give a proper json for opening the SummaryProductFragment.

![screenshot_2018-12-03-22-56-24-650](https://user-images.githubusercontent.com/32436867/49390587-65f78f00-f74f-11e8-8577-e9b3c3235546.jpeg)

shiv07tiwari:
@teolemon  Hello !
Can I work on this issue ?
If yes, what exactly do we need to show in the web page that we will open ?

jaindiv26:
@shiv07tiwari I'm already working on this feature.

shiv07tiwari:
Okay no issue.

PR #2012, opened 2018-12-03
Initial progress, for tweakable reference to dynamic product

jaindiv26:
## Description
@teolemon Here's just a sample how it would look, Do we also need to update this textView if not please give a suggestion about what would be the proper text for this, and also I need the API that would give a proper json for opening the SummaryProductFragment.

Issue Number: #2007 
 
 ## Screen-shots,

screenshot_2018-12-03-22-56-24-650


teolemon:
The idea is to make it a simple link to open in a webview, and to fetch the link from the server, if available, very much like the product count.
@stephanegigandet can you put up a file that would contain a short call to action wrapped by a <a> tag ?

jaindiv26:
@teolemon What if when we click on the textView instead of opening a webview, we open the SummaryProductFragment.

teolemon:
I want the system to be flexible. Being able to display an arbitrary webpage if we need it (product recall, blog post,…)
It's not really to display a single product.

jaindiv26:
@teolemon Please provide the appropriate text for the textView and the URL.

teolemon:
@jaindiv26 
https://world.openfoodfacts.org/files/tagline/tagline-en.json

jaindiv26:
@teolemon @Karljoones Here is the Screen Record of the changes I made. Please review.
https://drive.google.com/file/d/1BYASbIZOx5Y8arnA3mX28lLof9lX2-C5/view?usp=sharing

teolemon:
@stephanegigandet We need your input as to how we should query the server.

teolemon:
@jaindiv26 I've tweaked the JSON, based on suggestions by Stéphane:
https://static.openfoodfacts.org/files/tagline/tagline.json

- It's now multilingual
- It should try to match your exact Android locale (language + country)
- Fall back to at least your language
- Fall back to default
- Also, it shouldn't be clickable if no url is specified (I've included an example for Portuguese)

jaindiv26:
@teolemon Please give updates on this PR.

teolemon:
I got yesterday feedback from @stephanegigandet (see the comment just above) and I created a new multilingual JSON in response

jaindiv26:
Let me commit my changes.

app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/HomeFragment.java
```diff
@@ -192,4 +221,45 @@ public void onError(Throwable e) {
...
+
+    }
+
+    private void getTagline(){
+
+       Call<ArrayList<TaglineLanguageModel>> call = apiClient.getTagline();
+       call.enqueue(new Callback<ArrayList<TaglineLanguageModel>>() {
+           @Override
+           public void onResponse(Call<ArrayList<TaglineLanguageModel>> call, Response<ArrayList<TaglineLanguageModel>> response) {
+               if(response != null){
```
ripetrescu: You should check by **if (response.isSuccessful())** and handle it properly, otherwise the body might be null leading to NPE.

jaindiv26:
@stephanegigandet @ripetrescu All the requested changes are done. This is the screen recording, please review. https://drive.google.com/file/d/17itaMYGQbebKMkJBVNMzlXRCU283v--W/view?usp=sharing

teolemon:
Thank you so much @jaindiv26 for your patience during the back and forth
Thanks @ripetrescu for the review. Merging

PR #2014, opened 2018-12-04
Change the Allergen Fragment to show a message when empty

rahul-jha98:
##Description
The message which was supposed to be a popup is now embedded in the fragment view. It is visible when the list is empty and is hidden if the list has one element. According to me it was a better option as the view was quite empty when no items were present in the list.
Make sure that the code is fine. I have tried to use variable names as self-explanatory as possible.

#fixes 429 Transfer the allergen warning popup into the view

##Checklist
Simple UI change with few changes in code.
So satisfied almost all the checklist items


teolemon:
Merging. Thank you @rahul-jha98 and @brandenfung

PR #2040, opened 2018-12-30
hunger games for android

### LINKED ISSUE #2031: Add the category game in a webview accessible from the category explorer
image
Add the category game in a webview accessible from the category explorer. The banner doesn't need to be orange, it can be the same color as categories.

PrajwalM2212:
fixes #2031 

screenshot_1546142313


teolemon:
Note to self, change to the production URL before release.

PR #2070, opened 2019-01-07
Display CarbonFootprint in CardView and add Environment Tab

KarthikRIyer:
## Description

Carbon Footprint now displayed inside a CardView. 
Also added an Environment Tab to which Recycling Instructions will be added later.
 
whatsapp image 2019-01-08 at 2 41 58 am
whatsapp image 2019-01-08 at 2 41 38 am

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/product/ProductActivity.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/product/ProductActivity.java
@@ -50,6 +50,7 @@ import openfoodfacts.github.scrachx.openfood.views.customtabs.CustomTabActivityH
 import openfoodfacts.github.scrachx.openfood.views.customtabs.CustomTabsHelper;
 import openfoodfacts.github.scrachx.openfood.views.customtabs.WebViewFallback;
 import openfoodfacts.github.scrachx.openfood.views.listeners.OnRefreshListener;
+import openfoodfacts.github.scrachx.openfood.views.product.environment.EnvironmentProductFragment;
 import openfoodfacts.github.scrachx.openfood.views.product.ingredients.IngredientsProductFragment;
 import openfoodfacts.github.scrachx.openfood.views.product.nutrition.NutritionProductFragment;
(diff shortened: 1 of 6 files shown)


teolemon:
Can you move the Carbon footprint to the environment tab and hide the tab if it's not present?

PrajwalM2212:
@teolemon Is the environment tab only for OFF or even for OPF,OPFF and OBF ?

teolemon:
I would conditionnally show it on Open Food Facts, if and only if we have a carbon footprint.
Once we merge the PR for recycling instructions, I want to move recycling instructions and packaging in that tab.

app/src/main/java/openfoodfacts/github/scrachx/openfood/views/product/ProductActivity.java
```diff
@@ -264,6 +265,7 @@ private void setupViewPager( ViewPager viewPager )
 		{
 			adapterResult.addFragment( new NutritionProductFragment(), menuTitles[2] );
 			adapterResult.addFragment( new NutritionInfoProductFragment(), menuTitles[3] );
+			adapterResult.addFragment( new EnvironmentProductFragment(), "Environment" );
```
PrajwalM2212: If environment tab is OFF only the a guard has to be put here.

PrajwalM2212: You need to add this to `menuTitles` string array and then reference it here. Do the same for other entries that may have missed this logic.

KarthikRIyer: Do I need to add it manually to all the languages? Adding it to the default strings.xml doesn't seem to work

huzaifaiftikhar: @KarthikRIyer No, you don't need to add it manually to all the languages, just add the String to the default `string.xml` file. The translations are handled by Crowdin. You don't need to worry about that.

teolemon:
Thank you @KarthikRIyer 👍 
Thanks @PrajwalM2212 for the review and feedback

Issue #2073, opened 2019-01-09
Small issue with the store field


teolemon:
![image](https://user-images.githubusercontent.com/1689815/50897423-52a1a880-140c-11e9-9899-11072ace9463.png)
It seems that an extra space is introduced on the 2nd and 3rd value. It ain't like that on other fields.

teolemon:
Example barcode: 3245412051661

teolemon:
https://barcode.tec-it.com/en/EAN13?data=3245412051661

subhanjansaha:
I'll work on this.

teolemon:
it was fixed. closing

PR #2084, opened 2019-01-11
Fixed: Issue: #2073, Small issue in store and other fields of Overview Tab.

Shritesh99:
## Description

There is a leading whitespace in the store field labels so, I have just applied String.trim() function on label for removing leading whitespaces.

## Screen-shots, if any
 fixes
fixes copy


teolemon:
Thanks @PrajwalM2212 for the review
Thanks @Shritesh99 for the PR. Merging 👍

Issue #2096, opened 2019-01-13
Search doesn't work when invoked from the "Products to be completed" screen


teolemon:
Search doesn't work when invoked from the "Products to be completed" screen

subhanjansaha:
I'll work on this.

teolemon:
It should be fixed, but this needs to be verified.

teolemon:
This is fixed.

PR #2100, opened 2019-01-15
Fix search

PrajwalM2212:
#2096

When  `Products to be completed`  was opened and when we typed say `cola`, the app would search for `cola` in `Products to be completed api`. So only few or no results were obtained.  This would be same for any other `search type`.  

But now I have made so that any new query will default to `seach api`.

fix_ssearch

@teolemon


teolemon:
looks good, merging

Issue #2086, opened 2019-01-11
Trash icon visible after swiping in History section


tronku:
**Summary:** 

In the history section, the trash icon used to delete an item is visible even after deleting the item.

**Steps to reproduce:** 

Go to History section from Navigation Drawer. Swipe any item to the left and click the trash icon visible on the right. The item will get deleted but the trash icon will be still visible.

**Expected behavior:** 

After clicking the trash icon, it should be gone.

**Observed behavior:** 

The trash icon found by swiping the item to the left is visible even after deleting the item.

**Device and Android version:** 

Model - Mi Redmi Note 5
Android version - Android Oreo (8.1)
ROM - Custom

**Screenshots:** 

![whatsappvideo20190111at84](https://user-images.githubusercontent.com/31654207/51043670-4290f100-15e5-11e9-85a8-36699bb58794.gif)

Shritesh99:
Working on it.

PR #2104, opened 2019-01-16
Fixed: issue: #2086, Trash icon visible after swiping in History section.

Shritesh99:
## Description

Trash icon visible after swiping in History section. After, figuring out the the why trash icon is visible after right swipe, I saw that it is visible because the whatever ButtonShowedState is, it is visible. So I just fixed the icon visibility when button ButtonShowedState is `RIGHT_VISIBLE`.

Issue #2086
 
 ## Screen-shots, if any
 
whatsappvideo20190116at62


Shritesh99:
Check out my latest commit, tried a different approach in it.

PR #2113, opened 2019-01-17
Display useful message

### LINKED ISSUE #2110: Replace log message in Product Photos by something more useful
image
The success message is very geeky.

PrajwalM2212:
fixes #2110


app/src/main/res/values/strings.xml
```diff
@@ -900,6 +900,7 @@
     <string name="set_img_front">Set Image as Front</string>
     <string name="report_img">Report Image</string>
     <string name="detected_ingredients">Detected Ingredients</string>
+    <string name="set_image_name">"Successfully set the image as "</string>
```
teolemon:
```suggestion
    <string name="set_image_name">Successfully set the image as </string>
```

PrajwalM2212: @teolemon I had to do that for adding the extra space adter `as`. The double quotes will not be rendered in the app.

teolemon:
@ripetrescu your issue was solved. Looks good ?

app/src/main/res/values/strings.xml
teolemon: Translators don't have to question themselves about double quotes @PrajwalM2212

PrajwalM2212: @teolemon I did not think about from that perspective but it is fixed now

PR #2313, opened 2019-02-26
Fixes #2291: Deduplicate the code used in the slideup view and the regular product view

### LINKED ISSUE #2291: Deduplicate the code used in the slideup view and the regular product view
Deduplicate the code used in the slideup view and the regular product view.
This is 90+% the same code, and it's causing slipups when one is changed and not the other.

subhanjansaha:
## Description

The large methods in `ProductActivity` were converted to static methods which can be referenced from `ProductFragment`. Suitable modifications were mode to get the static counterparts to work but as a result 142 unnecessary lines of code were eliminated from `ProductFragment` and now the changes made in the static methods would be reflected everywhere.

#fixes #2291


Karljoones:
Merged, thanks @subhanjansaha

PR #2392, opened 2019-03-10
Harmonize the size of the buttons in the summary

teolemon:
Harmonize the size of the buttons in the summary (yes still very ugly)
image

image


teolemon:
This breaks the last word of compute Nutri-Score prompt

teolemon:
I've gone a bit further
image

PR #2419, opened 2019-03-13
Fix #2391 text "Add nutrition facts to compute the Nutri-Score" should be translatable

### LINKED ISSUE #2391: This should be translatable
https://github.com/openfoodfacts/openfoodfacts-androidapp/blob/fa54c4b2a6f69ae2590387f2b02a63e73d07b5f3/app/src/main/res/layout/fragment_summary_product.xml#L232

deniger:
## Description
+ Replace the text "Add nutrition facts to compute the Nutri-Score" by an id
+ add related entry in strings.xml files

#fixes 2391


teolemon:
Looks good to me. Thanks @deniger :-)

Issue #2374, opened 2019-03-08
Nutrition units incorrectly managed


verdy-p:
While entering nutrition info, it forces the input for 100g even if we indicate a portion. The conversion bet, even kilojoules and kilo calories are incorrect. Units in milligrammes for example vitamins are recorded but incorrectly displayed divided by 1000, as if they were in grams, but displayed still as milligrams...

The results is completely unreliable. Does not match the actual input. 

Notations are then incorrectly computed!!

Please fix the measurement units in the Android app!!!

teolemon:
Thanks a lot @verdy-p 
Are you using the regular version or the beta ?

verdy-p:
The regular Android App; this is not the beta.

Le ven. 8 mars 2019 à 15:24, Pierre Slamich <notifications@github.com> a
écrit :

PR #2453, opened 2019-03-21
fix #2403, #2374 Nutrient name should always be visible in edit mode (even if we have a value)

### LINKED ISSUE #2403: Use TextInputLayout in all fields of the edit form
https://developer.android.com/reference/android/support/design/widget/TextInputLayout

deniger:
+ Use TextInputLayout to always display field name
+ Clean the fragment used to edit nutrient values by removing duplicate code
+ fix issues with value validation for nutrient

## Description

Describe the changes made and why they were made instead of how they were made.

#fixes 2403, 2374
 
 ## Screen-shots, if any
 
 ## Checklist
 
 - [ ] Apply the `AndroidStyle.xml` style template to your code in Android Studio.
 - [ ] Code is well documented
 - [ ] Include unit tests for new functionality
 - [ ] All user-visible strings are made translatable
 - [ ] Code passes Travis builds in your branch
 - [ ] If you have multiple commits please combine them into one commit by squashing them.
 - [ ] Read and understood the contribution guidelines .


teolemon:
- There are leftovers from the URL field in simplified mode (barcode and magnifying glass)
- The fields are not the same height
- A bit too much left margin for the field names, and perhaps not enough top margin as compared to the field

image

deniger:
I will have a look tomorrow.
For the left margin of the field name: it s aligned with the text field edit zone ( the cursor). I can modify it but will have to modify the left padding of the edit text.



---- On Thu, 21 Mar 2019 17:06:53 +0100 notifications@github.com wrote ----


There are leftovers from the URL field in simplified mode (barcode and magnifying glass)
The fields are not the same height
A bit too much left margin for the field names, and perhaps not enough top margin as compared to the field

deniger:
I review the points:

image

Note: I move the 2 images (scan and search) at the bottom of the link textfield. Is it ok for you ?

teolemon:
I would have hidden completely the link field rather than showing it. The minimal mode aims at letting you compute the Nutri-Score and NOVA groups.

deniger:
I'm not sure I get it all:)
How do you activate the minimal mode? Are we ok that in normal mode the link field should be visible ?
I only made tests in normal mode but in the code or seems that link field is hidden in minimal mode.


---- On Sat, 23 Mar 2019 17:02:30 +0100 notifications@github.com wrote ----



I would have hidden completely the link field rather than showing it. The minimal mode aims at letting you compute the Nutri-Score and NOVA groups.

teolemon:
It's now the default mode, and the link should not be visible.

Issue #2454, opened 2019-03-21
Nutrient using % DV as unit are not displayed correctly


deniger:
To check it, open the product "1" ( Golden Virgina). The zinc value is not correctly displayed.
The correct value shoud be 444 %DV
The issue:
the api give values in gram but the application doesn't take it in account.

deniger:
Fixed by #2455

PR #2455, opened 2019-03-22
fix #2374, #2454 Nutrition units incorrectly managed

### LINKED ISSUE #2374: Nutrition units incorrectly managed

deniger:
+ fix conversion issues

#fixes #2374, #2454


deniger:
There is an issue with travis:
https://travis-ci.org/openfoodfacts/openfoodfacts-androidapp/jobs/509975888
it seems that oracle jdk can't be installed...

teolemon:
```Caused by: com.android.builder.errors.EvalIssueException: Failed to install the following Android SDK packages as some licences have not been accepted.
   build-tools;28.0.3 Android SDK Build-Tools 28.0.3
To build this project, accept the SDK license agreements and install the missing components using the Android Studio SDK Manager.
Alternatively, to transfer the license agreements from one workstation to another, see http://d.android.com/r/studio-ui/export-licenses.html```

teolemon:
mmmh…
https://github.com/openfoodfacts/openfoodfacts-androidapp/blob/master/.travis.yml

deniger:
A problem occurred configuring project ':app'.
> Failed to install the following Android SDK packages as some licences have not been accepted.
     build-tools;28.0.3 Android SDK

deniger:
In Travis I suppose we should change 28.0.2 to 28.0.3 or maybe there is a way to not specify this version

teolemon:
yes indeed

teolemon:
@deniger small conflict to solve

teolemon:
image
image

teolemon:
image
image

deniger:
Currently these 2 results are normal (without the taxonomy). On the Android app, we use the unit entered by the user to display nutrient values. For fat, mg has been used. for vitamin K,  % DV has been used but in the api gram are returned and so we display the result in gram....this PR will only solve the mismatches in unit display ( gram as milligramme) and we should load the taxonomy to have a correct display

deniger:
This PR is just to fix this kind of issue :). On the current app, unit are mixed and for Golden Viginia, the value "3.5 mg" is displayed instead of 3.5 g ( or 3.500 mg in this PR)
image

teolemon:
Ok, it doesn't match the website, but it's factually correct. Merging. Thanks @deniger

PR #2492, opened 2019-03-28
Add Uri null check for share image intent (#2444)

### LINKED ISSUE #2444: java.lang.NullPointerException:    at openfoodfacts.github.scrachx.openfood.views.MainActivity.detectBarCodeInImage
```
Caused by: java.lang.NullPointerException: 
  at openfoodfacts.github.scrachx.openfood.views.MainActivity.detectBarCodeInImage (MainActivity.java)
  or                     .lambda$myContributions$5$MainActivity (MainActivity.java)
  at openfoodfacts.github.scrachx.openfood.views.MainActivity.access$000 (MainActivity.java)
  or                     .changeFragment (MainActivity.java)
  or                     .chooseDialog (MainActivity.java)
  or                     .createAlertDialog (MainActivity.java)
  or                     .lambda$myContributions$6$MainActivity (MainActivity.java)
  or                     .lambda$onCreate$0$MainActivity (MainActivity.java)
  or                     .lambda$onCreate$3$MainActivity (MainActivity.java)
  or                     .lambda$scan$4$MainActivity (MainActivity.java)
  or                     .setItemSelected (MainActivity.java)
  at openfoodfacts.github.scrachx.openfood.views.MainActivity.handleSendImage (MainActivity.java)
  or                     .lambda$null$1$MainActivity (MainActivity.java)
  at openfoodfacts.github.scrachx.openfood.views.MainActivity.handleIntent (MainActivity.java)
  or                     .lambda$null$2$MainActivity (MainActivity.java)
  or                     .updateBadgeOfflineEditDrawerITem (MainActivity.java)
  at openfoodfacts.github.scrachx.openfood.views.MainActivity.onCreate (MainActivity.java)
  at android.app.Activity.performCreate (Activity.java:6910)
  at android.app.Instrumentation.callActivityOnCreate (Instrumentation.java:1123)
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:2746)
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:2864)
  at android.app.ActivityThread.-wrap12 (ActivityThread.java)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1567)
  at android.os.Handler.dispatchMessage (Handler.java:105)
  at android.os.Looper.loop (Looper.java:156)
  at android.app.ActivityThread.main (ActivityThread.java:6523)
  at java.lang.reflect.Method.invoke (Method.java)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:942)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:832)
```
image

tomdaly:
## Description

If app is opened from a multiple image share intent and an image has
invalid Uri (null), this check fixes openInputStream acting on null

`getContentResolver.openInputStream(uri)` throws a NullPointerException if the Uri is null (invalid), so a null check is made at start of Uri list iteration

Problem exists as single image check in `handleSendImage` does check for null, but `handleSendMultipleImages` does not check individual list entries for null so modifications were made to ensure null check occurs in every instance

fixes #2444

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/MainActivity.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/MainActivity.java
@@ -73,6 +73,7 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.InputStream;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Hashtable;
 import java.util.Objects;
 
@@ -899,7 +900,6 @@ public class MainActivity extends BaseActivity implements CustomTabActivityHelpe
         Uri selectedImage = null;
         ArrayList<Uri> selectedImagesArray = new ArrayList<>();
         selectedImage = intent.getParcelableExtra(Intent.EXTRA_STREAM);
-        boolean isBarCodePresent = false;
         if (selectedImage != null) {
             selectedImagesArray.add(selectedImage);
             chooseDialog(selectedImagesArray);
@@ -909,6 +909,7 @@ public class MainActivity extends BaseActivity implements CustomTabActivityHelpe
     private void handleSendMultipleImages(Intent intent) {
         ArrayList<Uri> selectedImagesArray = intent.getParcelableArrayListExtra(Intent.EXTRA_STREAM);
         if (selectedImagesArray != null) {
+            selectedImagesArray.removeAll(Collections.singleton(null));
             chooseDialog(selectedImagesArray);
         }
     }


app/src/main/java/openfoodfacts/github/scrachx/openfood/views/MainActivity.java
```diff
@@ -899,11 +899,8 @@ private void handleSendImage(Intent intent) {
         Uri selectedImage = null;
         ArrayList<Uri> selectedImagesArray = new ArrayList<>();
         selectedImage = intent.getParcelableExtra(Intent.EXTRA_STREAM);
-        boolean isBarCodePresent = false;
-        if (selectedImage != null) {
-            selectedImagesArray.add(selectedImage);
-            chooseDialog(selectedImagesArray);
-        }
+        selectedImagesArray.add(selectedImage);
```
deniger: Why did you remove the check if (selectedImage != null) {....
Is it not usefull ?

app/src/main/java/openfoodfacts/github/scrachx/openfood/views/MainActivity.java
```diff
@@ -926,6 +923,7 @@ private void chooseDialog(ArrayList<Uri> selectedImagesArray) {
     private boolean detectBarCodeInImage(ArrayList<Uri> selectedImages) {
         InputStream imageStream = null;
```
deniger: To do in another PR maybe: the InputStream is not closed ( memory leak...)

deniger:
Here is a related issue ( clean up): https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/2493

app/src/main/java/openfoodfacts/github/scrachx/openfood/views/MainActivity.java
tomdaly: Could leave it in, just removed it as running chooseDialog() then runs detectBarCodeInImage(), and the first line of that is the added 'if (uri == null)' so it checks for null anyway

Maybe I should leave it in to check for null before so it doesn't waste time getting to the new null check

deniger: if "selectedImage" is null, chooseDialog could be launched with a null value, then `product_images.setAdapter(new PhotosAdapter( uri));` (line 983) will, then `PhotosAdapter#onBindViewHolder`
But I'm not sure it has an impact....

tomdaly: Ah I see why that would be a problem - I will add that back in, but if one of the Uris in the ArrayList is null (because it's invalid) after `handleSendMultipleImages` is run, that line (`MainActivity#983`) will create a `PhotosAdapter` with an ArrayList containing a null value - if `PhotosAdapter#onBindViewHolder` is called then will it also cause a NullPointerException at `PhotosAdapter#40` (`load(mPhotos.get(position)`).

To solve all of these problems, I could re-add the null check in `handleSendImage`  and at the start of `handleSendMultipleImages` iterate over the ArrayList to remove null values; this would mean the new `if (uri == null) return false` line can be removed

What do you think?

deniger: 👍  Ok with your last suggestion 
Thanks

PR #2532, opened 2019-04-08
travis optimization

deniger:
+ run unit tests before android tests ( check that the build failed if a unit test failed)
+ use cache as mentionned in https://docs.travis-ci.com/user/languages/android/#caching
+ removed unusefull apk stored on github
+ deploy apk on the releases page https://github.com/openfoodfacts/openfoodfacts-androidapp/releases ( see https://github.com/deniger/openfoodfacts-androidapp/releases for examples/tests). apk will be suffixed by the branch name

WARN: only the branch will product APK and not the PRs.


teolemon:
Can we write the output APKs somewhere ? On the server for instance ?

deniger:
good point... I'll check this

Issue #2512, opened 2019-04-02
Don't allow user to create multiple product lists with same name


mustaqmustu:
**Summary:** 
Don't allow user to create multiple product lists with same name

**Screenshots:** 
![ezgif com-video-to-gif(10)](https://user-images.githubusercontent.com/43813438/55401714-3c206e80-556f-11e9-91d5-fb68ae07c3b3.gif)

StephaneC:
How do you want to manage backward compatibility on devices with multiples lists and already existed Database ?
I though about adding @Index on listName, but it will crash older app.  

If you want to manage that in a resilient way, I'll _just_ add a check on list name before adding. 
That's less maintainable.

PR #2534, opened 2019-04-08
 #2512 Multiple lists with same name.

StephaneC:
## Description

- Updated database schema version to add Unique on listName
- Add check on listName before adding to resolve backward compatibility
 
To do in the future : update MaterialDialog deps. 
Not done here, because too much work. OFF has been used 24 times. 
Need a dedicated Issue and a full test of the app. 

#fixes {2512}


teolemon:
Thanks @StephaneC 
I just relaunched the 2/4 travis builds.
@brandenfung can you have a look. There's a DB schema change.

teolemon:
@StephaneC it works well for me. The app is blocked on first use, but I don't think it's related (see my latest bug)

deniger:
Tests are ok.
Something I don't understand but it works :)
- I have 2 lists with the same name
- I use this fix
- I suppose that some issues should appear with DB as the name is not unique
- But it works...

So ok for me

PR #2597, opened 2019-05-02
Fix #2551 by making app compatible with android 4.2

### LINKED ISSUE #2551: Crash when scanning
[logcat.txt.](https://github.com/openfoodfacts/openfoodfacts-androidapp/files/3067390/logcat.txt)

**Summary:** 

Tried with 3 different products, crash with all 3
Tried with the same version on a  Samsung SM-G361F ( Android 5.1.1 ) --> doesn't crash

Same apparent behavior (crash) with the previous beta version 3.0.7 ( sorry no traces )

Samsung GT-I9195
Android 4.2.2

deniger:
Warnings for future dev
- Apache method called StringUtils#join should not be used as it uses Objects class.
- drawableLeft with svg assets should be avoided and replaced  by  setCompoundDrawablesWithIntrinsicBounds (a better solution could be to use https://developer.android.com/jetpack/androidx/releases/ but weird issue with it..)

Tests:
- Emulatior with Android 4.2
- Search product
- Edit product

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,6 @@ cache:
     # Gradle dependencies
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
-    - $HOME/.android/build-cache
 
     # Cache maven
     - $HOME/.m2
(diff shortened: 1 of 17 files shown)


teolemon:
I'm merging and releasing.

PR #2638, opened 2019-05-14
extract string from layout

deniger:
extract string from layout


app/src/main/java/openfoodfacts/github/scrachx/openfood/views/ContinuousScanActivity.java
```diff
@@ -324,7 +329,8 @@ public void onError() {
                         fragmentTransaction.replace(R.id.frame_layout, productFragment);
                         fragmentTransaction.setTransition(FragmentTransaction.TRANSIT_FRAGMENT_FADE);
                         fragmentTransaction.commit();
-                        showFirstScanTooltipIfNeeded();
+
+
```
huzaifaiftikhar: You can reformat the code and remove these extra blank lines

deniger: yep, will do

PR #2667, opened 2019-05-27
fix issue with offline view + scan bottom view + make OpenFoodAPIClient more robust

deniger:
+ Fix issue with offline view ( text not visible)
+ make OpenFoodAPIClient more robust
+ try to fix sonarcloud connection problem


deniger:
If the user enter a barcode with keyboard, the bottom view didn't use all the required height

PR #2674, opened 2019-05-28
Fixes #2673: Replaced SharedPreferences commit() method with apply()

### LINKED ISSUE #2673: Use apply() on SharedPreferences instead of commit()
**Summary:** 

Presently SharedPreferences is using commit() method to write its data on SharedPreferences in the below code:

                        editor = pref.edit();
                        editor.putBoolean(IS_FIRST_TIME_LAUNCH, isFirstTime);
                        editor.commit(); 

The commit() method should be replaced with apply() method because commit blocks and writes its data to persistent storage immediately, apply will handle it in the background. So commit() should be replaced with apply() on SharedPreferences.

RajaVamsi11:
## Description
Replaced SharedPreferences commit() method with apply() which will handle writing data in the data storage in the background.

Fixes #2673


teolemon:
Thanks @RajaVamsi11 :-) 
Thank you @deniger for the review 👍

Issue #2703, opened 2019-06-03
static field ProductFragment.productState should be removed


deniger:
+ This field is set each time a product is opened
+ it's used in BaseFragment#getStateFromActivityIntent :(
+ The full workflow should be revisited to remove this field

PR #2704, opened 2019-06-03
2 minor changes for sonarqube reviews

deniger:
+ Add 2 comments for sonarqube issues
+ add a log if a file can't be deleted

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/product/ProductFragment.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/product/ProductFragment.java
@@ -43,7 +43,7 @@ import static android.app.Activity.RESULT_OK;
 public class ProductFragment extends Fragment implements OnRefreshListener {
 
     private static final int LOGIN_ACTIVITY_REQUEST_CODE = 1;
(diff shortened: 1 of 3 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/views/product/ProductFragment.java
```diff
@@ -43,7 +43,7 @@
 public class ProductFragment extends Fragment implements OnRefreshListener {
 
     private static final int LOGIN_ACTIVITY_REQUEST_CODE = 1;
-    public static State productState;
+    public static State productState;//NOSONAR To be changed ASAP !
```
teolemon: perhaps adding a TODO ?

deniger: I created an issue about this point:
https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/2703

teolemon: then, LGTM, I'm merging

PR #2731, opened 2019-06-09
[Translating and new Gradle Version]

pvergara:
## Description

* Adding/Updating new Galician translation
* Updating the gradle version


VaiTon:
Translations should be made on http://translate.openfoodfacts.org so we can track them

pvergara:
Ok thnx for the review, I'll close the pull request.

PR #2815, opened 2019-07-03
Scan activity: add a dialog for allergen

### LINKED ISSUE #2809: Handle plural singular in additive count
image

deniger:
+ if the user added allergens, a dialog should be displayed in the scan activity.
+ Fix #2809


app/src/main/java/openfoodfacts/github/scrachx/openfood/views/adapters/HistoryListAdapter.java
```diff
@@ -134,7 +134,7 @@ public void calcTime(Date date, HistoryScanHolder holder) {
         long hours = TimeUnit.MILLISECONDS.toHours(now.getTime() - date.getTime());
         long days = TimeUnit.MILLISECONDS.toDays(now.getTime() - date.getTime());
 
-        String secText = String.valueOf(seconds) + " seconds ago";
+        String secText = seconds + " seconds ago";
```
teolemon: This should be translatable

deniger: done

PR #2834, opened 2019-07-12
Scan View: better management to display allergens infos.

deniger:
Scan UI Revisited:
+ remove blue button used to edit product
+ add info ( slide up...) as text
+ Allergens info displayed "inline" instead of using dialog.
image

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
@@ -33,6 +33,7 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.view.inputmethod.InputMethodManager;
 import androidx.annotation.DrawableRes;
+import androidx.annotation.NonNull;
 import androidx.appcompat.content.res.AppCompatResources;
 import androidx.browser.customtabs.CustomTabsIntent;
 import androidx.core.app.ActivityCompat;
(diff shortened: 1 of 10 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/views/ContinuousScanActivity.java
```diff
@@ -143,7 +140,7 @@ public void barcodeResult(BarcodeResult result) {
             }
 
             lastText = result.getText();
-            if (!(isFinishing() || isDestroyed())) {
```
teolemon: why did we add isDestroyed ? Why are we removing it ?

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
```diff
@@ -660,13 +665,13 @@ public static String getEnergy(String value) {
     }
 
     private static int convertKjToKcal(double kj) {
-        return kj != 0 ? Double.valueOf(kj / 4.1868d).intValue() : -1;
+        return kj != 0 ? (int) (kj / 4.1868d) : -1;
```
teolemon: I remember we had issues at some point with the API returning int and doubles at random. Not sure if fully fixed on all products @stephanegigandet @deniger

app/src/main/java/openfoodfacts/github/scrachx/openfood/views/ContinuousScanActivity.java
deniger: I remove it because it's not compatible api level 16
Added to avoid a bug when the barcodeResult method is called and activity is closed
I supposed isFinishing is doing the job

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
deniger: this modification  is just done to avoid boxing double to Double then to int

teolemon:
ok, merging 👍

PR #2850, opened 2019-08-01
Fixes #2844: Hiding the product card on pressing the scan button

### LINKED ISSUE #2844: Scan button in bottom menu should close product card
**Is your feature request related to a problem? Please describe.**
After scanning a product and expanding the card to full view, the only way to return to the scan view is by swiping the card down from the top. Alternatively, tapping back returns to the "welcome screen" and I have to tap scan again. Tapping scan should return to scan mode immediately.

**Describe the solution you'd like**
The scan button on the bottom nav should minimize or close the card and return to scan view.

image

sank20:
## Description
In the barcode scanning activity if the product card/ bottom sheet is visible on the screen, it should hide if the scan button is pressed for the next scan.

#fixes {2844}
 
 ## Screen-shots, if any


deniger:
Thanks @sank20

PR #2892, opened 2019-09-25
Fix duplicated lines and codes smells

deniger:
- Fix small issues with static imports
- Use Log instead of printstracktrace method
- refactor class YourListedProduct and HistoryScanActivity to avoid duplication.


teolemon:
Looks good to me. I think we can merge 👍

PR #2894, opened 2019-09-28
Change theme and fix bottom navigation issues

Steinhagen:
## Description
This commit introduces a number of fixes for the bottom navigation:
1. **Fix menu selection based on the correct activity.** This fix uses setSelectedItemId to make sure that the correct menu item is selected. This API was introduced in SDK 25.3.0.
2. **Change bottom navigation color scheme for usability reasons.** The color change was done to be identical to the one used by the Swift OFF application.
3. **Removed duplicate bottom navigation style declaration from the scan activity.**
4. **Made the active/inactive navigation item size identical.** This was more of a design change rather than a fix for an issue.
5. **Made sure that the Home navigation item is highlighted on activity resume.** Based on the current code base, this should be the only activity that requires this change.
6. **Made the sure that the Comparison activity will send you back to the Home activity when pressing the back button**. This is the default behaviour for all other activities.
7. **Rebased most files from this pull request to master.** This was needed to merge it.
8. **Modified the Scan activity based on the master xml.** The changes introduced by the third patch weren't all needed in that specific form. This was a rebase + tweak for that commit based on the current master branch.
9. **Removed the SDK restriction introduced by the first patch and tweaked the navigation selection logic.** Unfortunatelly, moving the bottom navigation activity highlight to the BottomNavigationListenerInstaller is not without masive changes to the Scan activity implementation. As a result, an alternative implementation was taken into consideration until a rework will be done. 
10. **Fixed trucated navigation item.** This issue was observed when the navigation item text was to large to be displayed. This issue was fixed into the latest material beta release. To see this fix, change the displayed language to 'Romanian'.

Navigation bar aspect:
navigationbar
:


app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AllergensAlertFragment.java
```diff
@@ -90,6 +90,7 @@ public void onViewCreated(@NonNull final View view, @Nullable Bundle savedInstan
         productRepository = ProductRepository.getInstance();
         mDataObserver = new DataObserver();
         bottomNavigationView  = view.findViewById((R.id.bottom_navigation));
+        bottomNavigationView.getMenu().getItem(0).setCheckable(false);
```
deniger: Is it possible to put the line 
`bottomNavigationView.getMenu().getItem(0).setCheckable(false);`
in the method 
`BottomNavigationListenerInstaller.install(bottomNavigationView, getActivity(), getContext());`

instead of in the calling fragment and avoid repeat the line in all fragments

teolemon:
@Steinhagen Can you document with a screenshot ?

Steinhagen:
I will make another commit for necessary changes in BottomNavigationListenerInstaller. I will also try to remove the 25.3.0 API dependency if it's possible.

teolemon:
small merge conflict to solve. Also, I've made a Dark mode logo for Open Food Facts.

teolemon:
@Steinhagen can you rebase the branch ?

teolemon:
@deniger @Steinhagen this would be a nice PR to merge :-)

Steinhagen:
This is just a rebase to master (to fix conflicts). The changes for removing the API dependencies haven't been made yet.

Steinhagen:
I had to rethink a bit the logic for selecting the navigation item (is still a bit redundant, but only adding the clean in BottomNavigationListenerInstaller.install function won't work for the Scan Activity (as it's currently implemented).

P.S.: I'm not an Android developer, so code review is needed before merging this modification.

teolemon:
@ripetrescu @deniger I'd very much like to include that one in the Christmas release if works well.

VaiTon:
> 
> 
> @ripetrescu @deniger I'd very much like to include that one in the Christmas release if works well.

Bump.

PR #2899, opened 2019-10-06
Closes #2398 Re-enable notifications for pending offline products saved

### LINKED ISSUE #2398: Reenable Wifi Service the right way
https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/1583
WifiUploadReceiver.java, line 39

huzaifaiftikhar:
## Description
- Use `WorkManager` instead of service to send notifications periodically.
- Use `PeriodicWorkRequest` with a `retryInterval` of 60 minutes and `flexInterval` of 10 mins, this means that the user will receive at max one notification in 50 mins if there are any products saved offline and the user is connected to a network.
- Reformat imports in `MainActivity.java`
- Remove files `WifiUploadReceiver.java` and `UploadService.java` since it is not being used anymore.
- Remove permission `android.permission.FOREGROUND_SERVICE` which is not being used anymore.

Fixes #2398
 
 ## Screen-shots, if any
image

 
 ## Checklist
 
 - [x] Code is well documented
 - [ ] Include unit tests for new functionality
 - [x] All user-visible strings are made translatable
 - [x] Code passes Travis builds in your branch
 - [x] Read and understood the contribution guidelines.


deniger:
In the previous version, notification will be displayed as soon Wifi is found. With this new version, the check will be done every XXmin. @teolemon is it ok for you ?

huzaifaiftikhar:
@deniger We can no more use the previously implemented way of running a service in the background and listening for changes in the network. No service can run in the background in Android 8 and above (Battery Optimization by Google). I found this to be the aptest way of implementing this. The minimum interval time can be 15 mins which would ideally be like real-time, but I avoided being so aggressive in posting notifications to the user so kept it 60 mins.

deniger:
thanks @huzaifaiftikhar for this info.
I'll take a look this morning

app/src/main/java/openfoodfacts/github/scrachx/openfood/views/MainActivity.java
```diff
@@ -115,6 +147,13 @@ protected void onCreate(Bundle savedInstanceState) {
         }
         setContentView(R.layout.activity_main);
 
+        final PeriodicWorkRequest periodicWorkRequest = new PeriodicWorkRequest.Builder(OfflineEditPendingProductsWorker.class, 60, TimeUnit.MINUTES, 10, TimeUnit.MINUTES)
+            .setConstraints(new Constraints.Builder().setRequiredNetworkType(NetworkType.CONNECTED).build())
+            .addTag(OFFLINE_EDIT_PENDING_WORK_NAME)
+            .build();
+
+        WorkManager.getInstance(getApplicationContext()).enqueueUniquePeriodicWork(OFFLINE_EDIT_PENDING_WORK_NAME, ExistingPeriodicWorkPolicy.KEEP, periodicWorkRequest);
+
```
deniger: Is it possible to use ExistingPeriodicWorkPolicy.REPLACE instead of ExistingPeriodicWorkPolicy.KEEP.
The MainActivity can recreated: I added a breakpoint, use the application and noticed that the breakpoint can be hitted several times. So I guess that the work can be launched several times.

huzaifaiftikhar: Yes, the breakpoint must have hit several times but the work would only be created once since I have used [`enqueueUniquePeriodicWork`](https://developer.android.com/reference/androidx/work/WorkManager#enqueueUniquePeriodicWork(java.lang.String,%20androidx.work.ExistingPeriodicWorkPolicy,%20androidx.work.PeriodicWorkRequest)). Also the policy is kept as KEEP so that whenever this line is executed it doesn't replace the old work so that the timer is not refreshed again. Reference for [Policies](https://developer.android.com/reference/androidx/work/ExistingWorkPolicy.html).

PR #2900, opened 2019-10-06
Closes #2591 Add User-Agent Header in two network requests

### LINKED ISSUE #2591: Send header with tagline-off.json request
XXXX- - [29/Apr/2019:16:34:52 +0200] "GET /files/tagline/tagline-off.json HTTP/2.0" 200 424 "-" "okhttp/3.9.0"

Something like : ```Open Food Facts - Android client - version xxx ```

huzaifaiftikhar:
## Description
- Send User-Agent Header **Open Food Facts Official Android App 3.2.2** in the /tagline-xx.json and /1.json?fields=null GET requests.
- add `import openfoodfacts.github.scrachx.openfood.BuildConfig;` to `Utils.java` to use the correct BuildConfig, currently it was using the package `com.firebase.jobdispatcher` which was not desired and returned incorrect VersionName.
- Make User-Agent Header uniform throughout the app. They will look like:
`"<APP_NAME> Official Android App <VERSION> <TYPE>"`
`<Type>` is optional and the currently used values are `Scan` and `Search`.
All the User-Agent headers sent via the app would be of the above type only.
(Added a screenshot in a comment for reference)

fixes #2591 
 
 ## Screen-shots, if any
image

 ## Checklist
 
 - [x] Code is well documented
 - [x] Include unit tests for new functionality
 - [x] All user-visible strings are made translatable
 - [x] Code passes Travis builds in your branch
 - [x] Read and understood the contribution guidelines.

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/HomeFragment.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/HomeFragment.java
@@ -153,7 +153,7 @@ public class HomeFragment extends NavigationBaseFragment implements CustomTabAct
         super.onResume();
 
         int productCount = sp.getInt("productCount", 0);
-        apiClient.getTotalProductCount(getString(R.string.app_name) + " " + Utils.getUserAgent("OTHERS"))
+        apiClient.getTotalProductCount(Utils.getUserAgent())
             .subscribeOn(Schedulers.io())
             .observeOn(AndroidSchedulers.mainThread())
             .subscribe(new SingleObserver<Search>() {
@@ -224,7 +224,7 @@ public class HomeFragment extends NavigationBaseFragment implements CustomTabAct
 
     private void getTagline() {
         OpenFoodAPIService openFoodAPIService = new OpenFoodAPIClient(getActivity(), "https://ssl-api.openfoodfacts.org").getAPIService();
-        Call<ArrayList<TaglineLanguageModel>> call = openFoodAPIService.getTagline(getString(R.string.app_name) + " " + Utils.getUserAgent("OTHERS"));
+        Call<ArrayList<TaglineLanguageModel>> call = openFoodAPIService.getTagline(Utils.getUserAgent());
         call.enqueue(new Callback<ArrayList<TaglineLanguageModel>>() {
             @Override
             public void onResponse(Call<ArrayList<TaglineLanguageModel>> call, Response<ArrayList<TaglineLanguageModel>> response) {
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/HomeFragment.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/HomeFragment.java
@@ -153,7 +153,7 @@ public class HomeFragment extends NavigationBaseFragment implements CustomTabAct
         super.onResume();
 
         int productCount = sp.getInt("productCount", 0);
-        apiClient.getTotalProductCount()
+        apiClient.getTotalProductCount(getString(R.string.app_name) + " " + Utils.getUserAgent("OTHERS"))
             .subscribeOn(Schedulers.io())
             .observeOn(AndroidSchedulers.mainThread())
             .subscribe(new SingleObserver<Search>() {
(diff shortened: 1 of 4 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/HomeFragment.java
```diff
@@ -153,7 +153,7 @@ public void onResume() {
         super.onResume();
 
         int productCount = sp.getInt("productCount", 0);
-        apiClient.getTotalProductCount()
+        apiClient.getTotalProductCount(getString(R.string.app_name) + " " + Utils.getUserAgent("OTHERS"))
```
deniger: We should use a common way to build the header.
For instance in openfoodfacts.github.scrachx.openfood.network.OpenFoodAPIService#getProductByBarcodeSingle we use only this info:
`client.getProductByBarcodeSingle(product.getCode(), fields, Utils.getUserAgent(Utils.HEADER_USER_AGENT_SEARCH))`

I suggest to remove `getString(R.string.app_name) + " "` and change Utils.getUserAgent to add a prefix if needed.

app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/HomeFragment.java
```diff
@@ -224,7 +224,7 @@ public void onCustomTabsDisconnected() {
 
     private void getTagline() {
         OpenFoodAPIService openFoodAPIService = new OpenFoodAPIClient(getActivity(), "https://ssl-api.openfoodfacts.org").getAPIService();
-        Call<ArrayList<TaglineLanguageModel>> call = openFoodAPIService.getTagline();
+        Call<ArrayList<TaglineLanguageModel>> call = openFoodAPIService.getTagline(getString(R.string.app_name) + " " + Utils.getUserAgent("OTHERS"));
```
deniger: samething here. The prefix should be added by `Utils.getUserAgent`

In method `Utils.getUserAgent`, we could use `BuildConfig.APPLICATION_ID` as prefix and not `Official Android App `

deniger: @teolemon  is it for you ? I don't know the impact for the analytics part ?
The user agent will be:
openfoodfacts.github.scrachx.openbeauty 3.2.2 THE_TYPE

deniger: There is something wrong in the method `getUserAgent`.
The parameter `type` is not totally used ( only to test HEADER_USER_AGENT_SCAN / HEADER_USER_AGENT_SEARCH)

I suggest to replace: 
> public static String getUserAgent(String type) {
        final String prefix = "Official Android App ";
        if (type.equals(HEADER_USER_AGENT_SCAN)) {
            return prefix + BuildConfig.VERSION_NAME + " " + HEADER_USER_AGENT_SCAN;
        } else if (type.equals(HEADER_USER_AGENT_SEARCH)) {
            return prefix + BuildConfig.VERSION_NAME + " " + HEADER_USER_AGENT_SEARCH;
        }
        return prefix + BuildConfig.VERSION_NAME;
    }


by 

> public static String getUserAgent(String type) {
>         return BuildConfig.APPLICATION_ID+" " + BuildConfig.VERSION_NAME+" "+type;
> }

@huzaifaiftikhar please could you change this method

huzaifaiftikhar: @deniger I will make the changes suggested by you but will also overload this method without an argument so that we can get the user agent without THE_TYPE if required (which is required in the current scenario). 
Also, instead of using `BuildConfig.APPLICATION_ID` I will add another BuildConfig variable with the APP_NAME. This would look much better than the application ID. So we would be having the same user agent everywhere in the code.

app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/HomeFragment.java
huzaifaiftikhar: Done

huzaifaiftikhar:
**Updated User-Agent**
The User-Agent would be uniform through the app now. `TYPE` is made optional now. Currently, we only have `Scan` and `Search` as `TYPE`, we can think of adding more if required later on. 
image

PR #2918, opened 2019-11-01
OPF fixes

teolemon:


VaiTon:
@teolemon what are these new strings for?

teolemon:
to override food specific strings in non food versions of the app

PR #2930, opened 2019-11-19
Fixed tests

VaiTon:
Added RxJava2CallAdapterFactory to every instance of Retrofit.


huzaifaiftikhar:
@VaiTon Will review the PR this weekend

app/src/test/java/openfoodfacts/github/scrachx/openfood/network/OpenFoodAPIServiceTest.java
```diff
@@ -55,6 +60,7 @@ public void setUp() {
         serviceWrite = new Retrofit.Builder()
             .baseUrl(APIUtils.GET_API)
```
huzaifaiftikhar: Should we use the production API for testing? Shouldn't we use the testing URL i.e. `https://world.openfoodfacts.net` host? Current implementation leads to the creation of real product in our prod db. https://world.openfoodfacts.org/product/0978020137962/coca3-auchan. This product is created by the `post_product()` test.

app/src/test/java/openfoodfacts/github/scrachx/openfood/network/OpenFoodAPIServiceTest.java
```diff
@@ -185,23 +191,30 @@ public void getProduct_notFound() throws Exception {
     @Test
     public void post_product() throws IOException {
         SendProduct product = new SendProduct();
+
         product.setBarcode("978020137962");
         product.setName("coca3");
         product.setBrands("auchan");
         product.setWeight("300");
         product.setWeight_unit("g");
         product.setLang("fr");
```
huzaifaiftikhar: Is this needed anymore except the barcode? We can set everything inside the `productDetails` Map directly.

VaiTon: The only useful call would be `product.getQuantity()` on line 205.

teolemon: Can we use Brand neutral values like "Product name" or "OFF brand" ?

PR #2996, opened 2020-01-03
behaviour change for traces

### LINKED ISSUE #2994: Traces field is missing on product creation
The field "Traces" is missing on the tab "Ingredients" when I add new product. For example, "add" page looks like this:
Screenshot_2020-01-02-22-23-29-672_org openfoodfacts scanner

While "edit" page already contains necessary field:
IMG_20200102_223326

Android: 5.1.1
App version: 3.2.3

VaiTon:
- Closes #2994 
- ...
### Screenshots
immagine


teolemon:
Please don't merge this yet. Enabling fast addition is a way to reduce complaints about too many fields to fill. Most people are intimidated by the complexity of product addition or edition. They don't care about simple vs full mode. 
If we show traces, we show traces. But telling them they can tweak settings is going to raise more questions, and not very useful for 99% of the userbase.

PR #3007, opened 2020-01-27
Fix #2975 ask logged in users to rate the app

### LINKED ISSUE #2975: Ask logged-in users to rate the app
image
- We should disable this if the package name is the package name used for F-Droid (assuming those users either don't have the PlayStore or don't want to use it)
- This is a picture showing best practises. We should adapt it to our own app and design.

harshlele:
## Description
Asks logged in users to either rate the app or give feedback if they have the app installed for at least a week. I have tested it on my own device by logging in and moving the date ahead by a week

#2975

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -476,12 +476,12 @@
     <string name="prefs_language_summary">Change language</string>
 
     <!-- Strings related to feedback dialog -->
-    <string name="user_enjoying_app">Are you enjoying this app?</string>
-    <string name="user_ask_show_feedback_form">I\'m sorry to hear that! Could you tell me what happened?</string>
+    <string name="user_enjoying_app">Are you enjoying your experience with Open Food Facts?</string>
+    <string name="user_ask_show_feedback_form">We\'re sorry to hear that! Could you tell us what happened?</string>
     <string name="user_ask_rate_app">Great! Let others know what you think of this app!</string>
     <string name="rate_app">Rate app</string>
     <string name="no_thx">No thanks</string>
-    <string name="feedback_form_url">https://forms.gle/SWnx6h6QSUHcmFMy7</string>
+    <string name="feedback_form_url" translatable="false">https://forms.gle/SWnx6h6QSUHcmFMy7</string>
 
     <!-- Scan history -->
     <string name="scan_history_drawer">History</string>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -475,6 +475,14 @@
     <string name="prefs_language">Language</string>
     <string name="prefs_language_summary">Change language</string>
 
+    <!-- Strings related to feedback dialog -->
+    <string name="user_enjoying_app">Are you enjoying this app?</string>
+    <string name="user_ask_show_feedback_form">I\'m sorry to hear that! Could you tell me what happened?</string>
+    <string name="user_ask_rate_app">Great! Let others know what you think of this app!</string>
+    <string name="rate_app">Rate app</string>
+    <string name="no_thx">No thanks</string>
+    <string name="feedback_form_url">https://forms.gle/SWnx6h6QSUHcmFMy7</string>
```
teolemon: This should not be translatable

app/src/main/res/values/strings.xml
```diff
@@ -475,6 +475,14 @@
     <string name="prefs_language">Language</string>
     <string name="prefs_language_summary">Change language</string>
 
+    <!-- Strings related to feedback dialog -->
+    <string name="user_enjoying_app">Are you enjoying this app?</string>
+    <string name="user_ask_show_feedback_form">I\'m sorry to hear that! Could you tell me what happened?</string>
```
teolemon:
```suggestion
    <string name="user_ask_show_feedback_form">We\'re sorry to hear that! Could you tell us what happened?</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -475,6 +475,14 @@
     <string name="prefs_language">Language</string>
     <string name="prefs_language_summary">Change language</string>
 
+    <!-- Strings related to feedback dialog -->
+    <string name="user_enjoying_app">Are you enjoying this app?</string>
```
teolemon:
```suggestion
    <string name="user_enjoying_app">Are you enjoying your experience with Open Food Facts?</string>
```

app/src/main/res/values/strings.xml
harshlele: u mean just the url right?

teolemon: yes

Issue #2942, opened 2019-12-16
Add an alcohol warning


teolemon:
Add an alcohol warning
"Excessive consumption of alcohol is harmful to health, to be consumed with moderation."
on products with: "categories", "en:alcoholic-beverages"

PR #3010, opened 2020-02-02
added alcohol warning for products with category en:alcoholic-beverages

harshlele:
## Description

Added an alcohol warning for products that have the category en:alcoholic-beverages

#2942


teolemon:
Can you provide a screenshot of the change @harshlele ?

harshlele:
Screenshot_1580993674

teolemon:
A couple of remarks:
- I would perhaps extract the category-specific logic to some file named CategorySpecificLogic.java (with the idea that it might grow overtime)
- I wouldn't put it in the same line as categories (but possibly at the end of the card)

harshlele:
Something like this? 
Screenshot_1581342538

teolemon:
looks better 👍

harshlele:
@teolemon ive made the changes

VaiTon:
> 
> 
> @VaiTon storing logic in the translation files sounds a bit dangerous. I only gave a feedback in visual terms, not in code design

You're right. I'll make the string a field and review the entire PR.

app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/CategoryProductHelper.java
```diff
@@ -0,0 +1,140 @@
...
+                    ProductBrowsingListActivity.startActivity(baseFragment.getContext(),
+                        category.getCategoryTag(),
+                        category.getName(),
+                        SearchType.CATEGORY);
+                }
+            }
+        };
+        spannableStringBuilder.append(category.getName());
+        spannableStringBuilder.setSpan(clickableSpan, 0, spannableStringBuilder.length(), SPAN_EXCLUSIVE_EXCLUSIVE);
+        if (!category.isNotNull()) {
```
VaiTon: @harshlele I don't understand this line. Could you explain why is there a double negation (! and isNotNull) and also why this null check?

harshlele: i didnt write that method, just moved it to the new file because it was related to the categories

PR #3042, opened 2020-03-15
Minors code fixes, taken from SonarCloud

VaiTon:


teolemon:
- removes string param for alcoholic beverages 👍

Issue #3062, opened 2020-04-14
Bottom panel bug


sashabrava:
You have a great update of app interface. However, there is a small problem on the bottom panel remained.
If I:
1. Open the app
2. Choose then "history" tab
3. Return back,

the bottom panel still highlights "History" button.

If I: 
1. Choose "Scan" tab
2. Return back,

the blue button highlight on bottom panel disappears.

**Screenshots**

![Screenshot_2020-04-11-17-35-40-993_org openfoodfacts scanner](https://user-images.githubusercontent.com/11160215/79220157-cc1c1800-7e53-11ea-9eb8-24dad16e9737.png)

**Smartphone:**
 - Device: Xiaomi Redmi 3
 - OS: Android 5.1.1
 - Version 3.2.9

VaiTon:
Fixed in #3064.

PR #3064, opened 2020-04-14
Fixed bottom panel bug

sashabrava:
In order to fix the issue #3062 I moved the method selectNavigationItem() (which selects necessary item on the bottom panel) from onCreate to onResume method, so that it will run both on Activity creation and after the user pushes 'back' button.  And it will fix the issue, when user has highlighted "history" button while returning back to MainActivity (and all other combinations).
Tested by running OBF debug apk on Android 5.1.1, Android 9 devices.


VaiTon:
Code good. I have to test it, especially the part about the ContinuousScanActivity.

VaiTon:
Tested, merging.

PR #3072, opened 2020-04-19
[WIP] Offline first product edition

philippeauriach:
This PR is a work in progress. It aims to provide an offline first product creation / edition workflow.

Main goal is to always use the actual "offline" persisting solution (instead of checking if network is there, or waiting for the network to fail to fall down to offline), and create a periodic background job that will perform the upload. 

Most of the methods used to upload data / pictures in `OfflineProductService` comes from the old `OfflineEditFragment`.


teolemon:
A preliminary list of issues:

Issues:
- There's no feedback of success, and as upload is really fast, and save is instant, it almost feels like something got wrong.
- On the offline product addition, it does update well. The card does not display all the data added like the brand.
- When editing a product added offline, I have this weird behavior that was there before: it allows you to slideup the card (with the tooltip “You can slide up the card for more details”), and displays an empty card. I would block the slideup in this case, and say "You can click on the product to edit it"
- Offline added product does not appear in the history 
- the offline section having disappeared, there are 2 bars one on the other in the side menu
- The background upload seems to work, but is making the calls twice: https://world.openfoodfacts.org/product/8002270436635/san-pellegrino

Testing (todo pierre):
- the background upload would need to be tested on older versions of Android, and on background agressive brands

raphael0202:
I spotted a regression during product update. Steps to reproduce:
- Go to product 3564700492659
- update the salt value from 0.8 to 0.83
- go back
- Click on "Save" ("Would you like to save the product?")
- A request is sent to OFF, but without the updated value

philippeauriach:
@raphael0202 I just tried to debug this issue, the value is indeed sent, but maybe not in the correct format ?
Cf following screenshot presenting the product map being sent. It seems the issue occurs with all nutriment values ?

Capture d’écran 2020-05-09 à 12 41 59

teolemon:
(Unrelated: I was convinced we were sending partial product maps, with only changed values.)

raphael0202:
Shouldn't the nutrient values be prefixed with _serving or _100g ? That's what is indicated in the documentation

philippeauriach:
@teolemon yes we are supposed to send only modified values, but the "diff" is not done yet for the nutriments.
@raphael0202 I have something like this in mind but could not find it in the api ? Do you have a link ? Also, I took the same code as what was used before for the upload, so I might be missing something (or the nutriment edition was not working on android but I guess we would have spot that earlier).
Will double check tomorrow.

teolemon:
https://documenter.getpostman.com/view/8470508/SVtN3Wzy?version=latest#56e5ddf2-950d-4364-91d0-c02a9f0484e2

teolemon:
Also, at this point, I never remember if the doc is right, so I'd ping @stephanegigandet because he favors one way over another to make the API call (@stephanegigandet see the above screenshot to see what we're currently doing)

philippeauriach:
I just tried to change the salt value using master, I have some bad news, the change is not reflected. Here is the payload sent: 
Capture d’écran 2020-05-10 à 11 53 04

So it appears to not be a regression but a bug already present in the previous version. Before fixing it in my branch, I just would be sure that there is an actual problem on the android version (if you could confirm the bug occurs on your side on master ?), and to be sure of the system we should use.

raphael0202:
Some data comes partially from the producer (Scanmark), so it is possible that the nutrition fields are not writeable through the API (in which case it is not a issue with the Android app).

philippeauriach:
Good catch @raphael0202 , it works fine with https://world.openfoodfacts.org/cgi/product.pl?type=edit&code=31797302562278 . Gonna get another coffee :)

raphael0202:
Another behavior to fix:
- scan a new product
- add the missing front image by clicking on the image logo in the product header
- once the photo is validated, we go back to the continuous scan activity, instead of staying on the product page

raphael0202:
There is also an issue with the product history, it seems I switch between 2 product history lists.

stephanegigandet:
Not sure if this is related to the new version: when I add a new product, the focus is on the "Expiration date" field, I don't see the top of the form.

Also, I think this field should be removed from the mobile app.

image

stephanegigandet:
It's not the expiration date field in fact, it's the period after of opening. Which we don't have on OFF (we do on OBF).

teolemon:
Thanks @stephanegigandet I could reproduce it, only on initial addition.

teolemon:
I have filed a separate issue for the Best Before issue: https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/3124

teolemon:
@philippeauriach reports the back to scan screen is in master. So I see no objection for merging, although I'd like this behaviour fixed before release (this is a regression vs production)

PR #3074, opened 2020-04-21
Unit tests for utils classes

VaiTon:


app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
```diff
@@ -394,17 +394,18 @@ public static Bitmap getBitmapFromDrawable(Context context, @DrawableRes int dra
...
-     * @return round value or 0 if the value is empty or equals to 0
+     * @return round value <b>with 2 decimals</b> or 0 if the value is empty or equals to 0
      */
     public static String getRoundNumber(String value) {
         if ("0".equals(value)) {
             return value;
         }
 
-        if (isEmpty(value)) {
+        if (value == null || value.length() == 0) {
```
philippeauriach, line 408: Why not use `TextUtils.isEmpty()` ? (native from android)

VaiTon: Because I can't seem to figure out why when i pass the string `""` it doesn't return true.

Issue #1464, opened 2018-04-18
Add credits for contributors (coders and translators)


teolemon:
It would be nice to give credits to everyone involved in coding and translating in the settings.

A little rolling list of code contributors :-)

PrajwalM2212:
I will work on it .

PrajwalM2212:
@teolemon Where should is add this and how should this appear

PrajwalM2212:
@teolemon  Github api can be used to get list of all code contributors .
Dont know how to get list of translators

Karljoones:
@PrajwalM2212 probably easier just to have a file in the app (assets) that can be read into a list, the contributors to the Android project can just append their name onto it in a PR

teolemon:
I've added what we do on the iPhone version.

PR #3082, opened 2020-05-02
Added icons to fragment summary product.

### LINKED ISSUE #2970: Add icons to summary fields and reduce their length
image

VaiTon:
Fixes #2970.
image


teolemon:
https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/1464 
label icon by Seanau under CC-BY
Award icon by samuel nudds under CC-BY
Map is standard material, right @VaiTon ?

stephanegigandet:
Just curious: where do we list the CC-BY authors of the icons?

teolemon:
in #1464

stephanegigandet:
It looks weird to have icons for some tags but not others, like in the screen capture above where additives are more to the left because they don't have icons. Which fields can be displayed here, besides additives, categories, labels and packaging codes?

stephanegigandet:
I think there's a choice to be made: either display the field name, or display icons, but not both, as it makes things very crowded.

If we do display icons only, then we need to make sure the icons can be easily understood. The "tag" icon does not really work well for categories of products, as "tags" are used on actual products. But I can't think of any icon that really works for categories. Maybe the "list" icon.

teolemon:
Usability surveys in general (remember seeing that for LibreOffice) show that text+icon provide the maximum ease of use for users.

stephanegigandet:
Agreed, icons + text is better  than icons only. The issue I have is that here the text is not "part" of the icon (it's not centered below it, it's not the only text next to it), and it feels weird to have both like that.

stephanegigandet:
I guess a better question might be: why do we need icons here? Many people will understand icons as being buttons that they can click and do an action. Here the icons are used as labels for the clicable links that are next to them. That could be very confusing.

teolemon:
To give a soul to the app. Those are illustrations

VaiTon:
We need to create an issue for the attribution.
cc @teolemon

teolemon:
It already exists (#1464 ) @VaiTon

PR #3089, opened 2020-05-05
#3086 sort nutrients collator

raychenon:
See ticket #3086

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AddProductNutritionFactsFragment.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AddProductNutritionFactsFragment.java
@@ -42,11 +42,13 @@ import org.apache.commons.lang.StringUtils;
 
 import java.io.File;
 import java.net.URI;
+import java.text.Collator;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 
@@ -756,7 +758,7 @@ public class AddProductNutritionFactsFragment extends BaseFragment implements Ph
     @OnClick(R.id.btn_add_a_nutrient)
     void addNutrient() {
         String[] nutrients = getResources().getStringArray(R.array.nutrients_array);
-        Stringi18nUtils.sortAlphabetically(nutrients);
+        Stringi18nUtils.sortAlphabetically(nutrients, Collator.getInstance(Locale.getDefault()));
 
         new MaterialDialog.Builder(activity)
             .title(R.string.choose_nutrient)
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AddProductNutritionFactsFragment.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AddProductNutritionFactsFragment.java
@@ -755,9 +755,12 @@ public class AddProductNutritionFactsFragment extends BaseFragment implements Ph
 
     @OnClick(R.id.btn_add_a_nutrient)
     void addNutrient() {
+        String[] nutrients = getResources().getStringArray(R.array.nutrients_array);
+        Arrays.sort(nutrients);
+
         new MaterialDialog.Builder(activity)
             .title(R.string.choose_nutrient)
-            .items(R.array.nutrients_array)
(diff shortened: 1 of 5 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AddProductNutritionFactsFragment.java
```diff
@@ -67,6 +70,7 @@
 import openfoodfacts.github.scrachx.openfood.utils.FileUtils;
 import openfoodfacts.github.scrachx.openfood.utils.ProductUtils;
 import openfoodfacts.github.scrachx.openfood.utils.QuantityParserUtil;
+import openfoodfacts.github.scrachx.openfood.utils.StringComparator;
```
VaiTon, line 73: StringComparator imported but never used?

app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AddProductNutritionFactsFragment.java
```diff
@@ -754,17 +758,20 @@ private boolean isDataPer100() {
 
     @OnClick(R.id.btn_add_a_nutrient)
     void addNutrient() {
+        String[] nutrients = getResources().getStringArray(R.array.nutrients_array);
+        // sort string by ignoring accents on letters. The sorted strings will keep their accents.
+        Arrays.sort(nutrients, Collator.getInstance(Locale.US));
```
VaiTon, line 763: I bet you wanted to use StringComparator. :P

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/StringComparator.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Locale;
+
+/**
+ * Created by Raymond Chenon on 05/05/20.
+ * Compare the strings ignoring the accents on letters ([a-z][A-Z])
+ */
+public class StringComparator implements Comparator<String> {
+
+    public static <T> void sort(T[] arrays) {
+        Arrays.sort(arrays, Collator.getInstance(Locale.US));
```
VaiTon, line 16: Change `Locale.US` to the current Android language: `Locale.getDefault()`.

app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AddProductNutritionFactsFragment.java
raychenon: Thanks for the review.
This PR is experimental with Collator. I deleted the Comparator

I pushed the changes with Collator to https://github.com/openfoodfacts/openfoodfacts-androidapp/pull/3089

app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AddProductNutritionFactsFragment.java
raychenon: Changed.

OK, so you closed the previous one.
I reported back the changes here.

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/StringComparator.java
raychenon: Let's keep `Locale.US`, because it strips all accents.

`Locale.getDefault()` will be unpredictable because each language can have its own rules for how strings and letters are sorted.

http://tutorials.jenkov.com/java-internationalization/collator.html

VaiTon: We should follow how each language would sort the items, don't you think?
@raychenon
cc. @teolemon

teolemon: That's basically why I suggested collator. My understanding is that people took the pain to do the sorting right in every single language.

raychenon: Ok, now the `Collator` has to be passed to set explicitly the Locale. Otherwise, it could be hard to debug due to `Locale.getDefault()`

Didn't notice any difference between Locale.US, Locale.FR (no Locale.ES).

raychenon:
The PR was merged, the issue can be closed.

VaiTon:
Already closed, see #3086.

PR #3113, opened 2020-05-15
Serving size input more accessible

VaiTon:


teolemon:
There's the case where we need to input > or <. That's why we had opted for a full keyboard.
We should perhaps have something like this:
image

VaiTon:
> 
> 
> There's the case where we need to input > or <. That's why we had opted for a full keyboard.
> We should perhaps have something like this:
> image

We don't need this for the serving, which is what this PR is about. I'm dealing with what you are saying in another branch, not pushed yet.

Issue #2449, opened 2019-03-19
Reformat Nutrition tab


teolemon:
![54566761-1709f880-49d2-11e9-9fca-c614857bb013](https://user-images.githubusercontent.com/1689815/54604537-44dd5480-4a47-11e9-9bc8-6ad6488113ac.png)

VaiTon:
I don't think that putting the serving size in that table is the right thing to do.
That's because if we don't have nutrition values the table is not displayed, and the serving size is not displayed too even if we do have it.

PR #3173, opened 2020-05-23
Should fix #3053 "portion vs. serving" wording issue

### LINKED ISSUE #3053: Fix the "portion vs. serving" wording issue
**Describe the bug**

In the OpenFoodFacts app's tab "Nutrition", using the English localization, there is a line saying "Per portion: 250 ml". This is inconsistent wording given that one line below the same thing is called "serving":

off-androidapp PortionL10n cropped

In addition, the line "Per portion: […]" is formatted like a title for the table below it. That's not its role, so this choice of style is confusing.

**Expected behavior**

Using the same word for the same thing. And using formatting that makes clear that "Per portion: […]" is a note and refers to the "Per serving" column just below. As a proposal, it may look like this:

off-androidapp PortionL10n Proposal

**Smartphone**

* Device: Asus Nexus 7
* OS: Android 6.0.1

VaiTon:
"portion vs. serving" wording issue #3053


teolemon:
Can you provide a screenshot of how you fixed it ?

teolemon:
See https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/2449

teolemon:
Small conflict to solve

PR #3174, opened 2020-05-24
Fixes #3058 : Make the requests faster and less network intensive by using fields parameter

### LINKED ISSUE #3058: Make the requests faster and less network intensive by using the ?fields= parameter
Currently we send the following query when clicking on a Brand: `https://world.openfoodfacts.org/brand/Veuve%20Clicquot/1.json`  **(6.66 KB for 1 product)**

We should ask instead:
`https://world.openfoodfacts.org/brand/Veuve%20Clicquot/1.json?fields=brands,product_name_fr,product_name,image_small_url,image_front_thumb_url,quantity,nutrition_grades_tags` **(417 Bytes for 1 product)**

If you multiply the difference by a larger number of products (as is often the case, here there's a spelling mistake on the brand, thus only 1 product), you get a **much larger size difference** that's at least (6,2KB * N) (actually more, since there are no ingredients not nutrients on this champagne).
This might not matter on 5G or 4G, but it does matter on EDGE, 3G, and when you're in a supermarket with very bad network.

**Notes:**
- I added both image_small_url,image_front_thumb_url but in truth we can keep only the one we actually use.
- I added both product_name_XX and product_name to be able to display the product name in the language of the app, if available.

This should be done for  **Products to be completed** and **Facets** (the pages that appear when you click a value on a product page). Search and Scan have already been optimized

Varsha-29:
Added query fields to speed up brand detail fetch process
Changes made in OpenFoodAPIClient and OpenFoodAPIService

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/network/OpenFoodAPIClient.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/network/OpenFoodAPIClient.java
@@ -87,7 +87,7 @@ public class OpenFoodAPIClient {
     private static OkHttpClient httpClient = Utils.HttpClientBuilder();
     private final OpenFoodAPIService apiService;
     private Context mActivity;
-    private final String fieldsToFetchProductDetail = String
+    private final String FIELDS_TO_FETCH_FACETS = String
         .format("brands,%s,product_name,image_small_url,quantity,nutrition_grades_tags", getLocaleProductNameField());
 
     public OpenFoodAPIClient(Activity activity) {
@@ -348,7 +348,7 @@ public class OpenFoodAPIClient {
     }
 
     public void getBrand(final String brand, final int page, final OnBrandCallback onBrandCallback) {
-        apiService.getProductByBrands(brand, page, fieldsToFetchProductDetail).enqueue(new Callback<Search>() {
+        apiService.getProductByBrands(brand, page, FIELDS_TO_FETCH_FACETS).enqueue(new Callback<Search>() {
             @Override
             public void onResponse(@NonNull Call<Search> call, @NonNull Response<Search> response) {
                 onBrandCallback.onBrandResponse(true, response.body());
@@ -552,7 +552,7 @@ public class OpenFoodAPIClient {
     }
 
     public void getProductsByCountry(String country, final int page, final OnCountryCallback onCountryCallback) {
-        apiService.getProductsByCountry(country, page, fieldsToFetchProductDetail).enqueue(new Callback<Search>() {
+        apiService.getProductsByCountry(country, page, FIELDS_TO_FETCH_FACETS).enqueue(new Callback<Search>() {
             @Override
             public void onResponse(@NonNull Call<Search> call, @NonNull Response<Search> response) {
                 if (response.isSuccessful()) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/network/OpenFoodAPIClient.java
```diff
@@ -346,8 +346,8 @@ public OpenFoodAPIService getAPIService() {
     }
 
     public void getBrand(final String brand, final int page, final OnBrandCallback onBrandCallback) {
-
-        apiService.getProductByBrands(brand, page).enqueue(new Callback<Search>() {
+        String fields = "brands,product_name_fr,product_name,image_small_url,image_front_thumb_url,quantity,nutrition_grades_tags";
```
VaiTon, line 349: Make it final.
```suggestion
        final String fields = "brands,product_name_fr,product_name,image_small_url,image_front_thumb_url,quantity,nutrition_grades_tags";
```

app/src/main/java/openfoodfacts/github/scrachx/openfood/network/OpenFoodAPIClient.java
```diff
@@ -346,8 +346,8 @@ public OpenFoodAPIService getAPIService() {
...
+        apiService.getProductByBrands(brand, page, fields).enqueue(new Callback<Search>() {
             @Override
             public void onResponse(Call<Search> call, Response<Search> response) {
```
VaiTon, line 352:
```suggestion
            public void onResponse(@NonNull Call<Search> call, @NonNull Response<Search> response) {
```

Varsha-29:
Hi @VaiTon , 
Thanks for reviewing it. I was not quite clear about the other facets. Could you list them?

VaiTon:
We need to see which fields we are using, so we need to check every call to the function i think.
@teolemon what do you think?

teolemon:
Normally, the facets show all exactly the same values, which are the ones listed for brands. @VaiTon @Varsha-29

teolemon:
product_name_fr should be product_name_XX where XX is the locale of the app

app/src/main/java/openfoodfacts/github/scrachx/openfood/network/OpenFoodAPIClient.java
```diff
@@ -87,6 +87,7 @@
     private static OkHttpClient httpClient = Utils.HttpClientBuilder();
     private final OpenFoodAPIService apiService;
     private Context mActivity;
+    private final String fieldsToFetchProductDetail = "brands,product_name_fr,product_name,image_small_url,image_front_thumb_url,quantity,nutrition_grades_tags";
```
VaiTon, line 90: It is a constant, so i would rename it to:
```suggestion
    private final String FIELDS_TO_FETCH_FACETS = "brands,product_name_fr,product_name,image_small_url,image_front_thumb_url,quantity,nutrition_grades_tags";
```

VaiTon: Also what @teolemon said, `product_name_fr` should be `product_name_xx` with `xx` being the phone locale.

VaiTon:
Thank you!

teolemon:
I'm merging. Congratulations on your first OFF PR @Varsha-29 🎉

Issue #3024, opened 2020-02-20
Use the new Robotoff Question auth mechanism to give credit to contributors


teolemon:
Use the new Robotoff Question auth mechanism to give credit to contributors instead of Robotoff
Send headers: [“Authorization”: “Basic \(base64Credentials)“] with username:password

VaiTon:
Fixed with 18435ca7b9f8bad4998d6c909c182e9c00e42cd5

PR #3179, opened 2020-05-25
Use the new Robotoff Questions auth mechanism

VaiTon:
Closes https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/3024.

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/repositories/ProductRepository.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/repositories/ProductRepository.java
@@ -1,10 +1,12 @@
 package openfoodfacts.github.scrachx.openfood.repositories;
 
 import android.content.SharedPreferences;
+import android.util.Base64;
 import android.util.Log;
 
 import com.squareup.picasso.Picasso;
 
+import org.apache.commons.lang.StringUtils;
 import org.greenrobot.greendao.AbstractDao;
 import org.greenrobot.greendao.database.Database;
 import org.greenrobot.greendao.query.WhereCondition;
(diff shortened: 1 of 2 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/repositories/ProductRepository.java
```diff
@@ -846,7 +896,21 @@ private Boolean tableIsEmpty(AbstractDao dao) {
...
+        // if the user is logged in, send the auth, otherwise make it anonymous
+        final SharedPreferences userPref = OFFApplication.getInstance()
+            .getSharedPreferences("login", 0);
+
+        final String user = userPref.getString("user", "");
+        final String pass = userPref.getString("pass", "");
+
+        if (StringUtils.isNotBlank(user) && StringUtils.isNotBlank(pass)) {
+            final String baseAuth = "Basic " + Base64.encodeToString(
+                (user + ":" + pass).getBytes(), Base64.DEFAULT);
```
huzaifaiftikhar, line 908: Can we move this to utils? We might be needing this at other places too!

teolemon:
@huzaifaiftikhar : https://docs.google.com/document/d/1IoDy0toQrrqtWHvDYp2rEVw84Yq1J0x2pt-0RGTm7h0/edit#heading=h.lbdrybrecxlg

app/src/main/java/openfoodfacts/github/scrachx/openfood/repositories/ProductRepository.java
teolemon: mmh, not likely, it's a different mechanism for auth than regular login

VaiTon:
@teolemon was still testing, don't know why the request is not authenticated.

VaiTon:
My bad, shoud've set it as a draft.

teolemon:
ok, +1 point for the branching strategy 🤣 @VaiTon

app/src/main/java/openfoodfacts/github/scrachx/openfood/repositories/ProductRepository.java
huzaifaiftikhar: Doesn't look any different it's the standard Basic Auth. Honestly, we should use this with all our API endpoints (provided that the server supports it for all the APIs) that require authentication rather than sending passwords as plaintexts from the app.

PR #3219, opened 2020-06-06
Fixes #3058 : Make the requests faster and less network intensive by using the ?fields= parameter

### LINKED ISSUE #3058: Make the requests faster and less network intensive by using the ?fields= parameter

Varsha-29:
Fixes #3058 : Make the requests faster and less network intensive by using the ?fields= parameter
Added fields parameter to `byState` and `byAdditives` method
Modified the corresponding unit tests


app/src/main/java/openfoodfacts/github/scrachx/openfood/network/services/OpenFoodAPIService.java
```diff
@@ -353,7 +353,7 @@
     Call<Search> byUnknownNutrient(@Path("UnknownNutrient") String unknownNutrient);
 
     @GET("additive/{Additive}.json")
-    Call<Search> byAdditive(@Path("Additive") String additive);
+    Call<Search> byAdditive(@Path("Additive") String Additive, @Query("fields") String fields);
```
VaiTon, line 356:
```suggestion
    Call<Search> byAdditive(@Path("Additive") String additive, @Query("fields") String fields);
```

PR #3223, opened 2020-06-07
design: updated icons for obf and off

VaiTon:
image
image
image


VaiTon:
@teolemon are you comfortable changing the off icon too?

stephanegigandet:
I very much like the OBF icon. I think it would be a great idea to make the OFF icon in the same style (with food products instead of beauty products).

teolemon:
Can we use that as an interim, while someone designs one ?

teolemon:
Ok, let's merge 👍

PR #3225, opened 2020-06-07
fix: changed product comment depending on flavor

VaiTon:


app/src/main/java/openfoodfacts/github/scrachx/openfood/network/OpenFoodAPIClient.java
```diff
@@ -133,11 +132,32 @@ private OpenFoodAPIClient(String apiUrl) {
      * @param login the username
      */
     public static String getCommentToUpload(String login) {
-        String comment = PRODUCT_API_COMMENT + " " + Utils.getVersionName(OFFApplication.getInstance());
+        StringBuilder comment;
+        switch (BuildConfig.FLAVOR) {
+            case OFFApplication.OBF:
+                comment = new StringBuilder("Official OBF Android app (for Open Beauty Facts)");
```
teolemon, line 138:
```suggestion
                comment = new StringBuilder("Official Open Beauty Facts Android app");
```
Something like this ?

Issue #1799, opened 2018-07-30
Shake to scan seems broken


teolemon:
Shake to scan seems broken
Open Pet Food Facts, latest version

ivanhercaz:
Hi @teolemon,

I was experimenting today with Open Food Facts. I enabled the shake option but I got an important bug: when this option is enabled it enters automatically in the scan mode and I can't exit, if I press the back button of my smartphone the app enters again in loop.

I thought "maybe if I close the app and enter again", but not, this didn't work, because the app enter again in the scan mode. It is very strange, because I didn't shake the mobile, and it happens when I have it above a table too.

I have to test it in another smartphones, but I got this error in a Moto X Play with Android 7.1.1.

I don't know if this is related with this issue, due you specified "Open Pet Food Facts, latest version", but I prefer to comment it here to avoid duplicated issues. If you consider it could be another issue, tell me and I will report it with images or even a video-gif of what happened to me.

I await your answer!

PR #3251, opened 2020-06-13
fix: set fragment from code in CategoryActivity

VaiTon:


teolemon:
The things moved around (sensor) is the shake to scan feature. It has stopped to work a while ago, and can probably be removed.

teolemon:
Just tested the branch, it works (Android 11)

VaiTon:
> 
> 
> The things moved around (sensor) is the shake to scan feature. It has stopped to work a while ago, and can probably be removed.

Could you open an issue so we keep track? I'm merging.

teolemon:
#1799

PR #3258, opened 2020-06-15
Sentry crash reporting

philippeauriach:
This PR integrates Sentry and auomatic crash reporting. For now, the only additional information sent with crashes is the "flavor" set as a tag on each crash report.

### DIFF
--- a/sentry.properties
+++ b/sentry.properties
@@ -1,0 +1,4 @@
+defaults.project=openfoodfacts-android
+defaults.org=openfoodfacts
+auth.token=4664a0f9a08e4b2facef49f778b82321993e7e8b0783439fa81f7741b36249b3
+#defaults.url=//TODO:
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -9,6 +9,7 @@ plugins {
     id("com.android.application")
     id("de.timfreiheit.resourceplaceholders.plugin")
     id("org.greenrobot.greendao")
+    id("io.sentry.android.gradle")
 }
 
 fun obtainTestBuildType(): String {
@@ -114,6 +115,9 @@ dependencies {
 
     // UI Component : Chips Input
     implementation("com.hootsuite.android:nachos:1.2.0")
+
+    // Crash analytics
+    implementation("io.sentry:sentry-android:2.1.6")
 
     // Unit Testing
     testImplementation("junit:junit:4.13")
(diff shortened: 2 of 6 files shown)


VaiTon:
I have a few question:
* Where are the logs sent? Do we have a private instance or are we using the shared one?
* Isn't this implemented already in the play store? cc @teolemon 
* If the first one is true, what do we know about user privacy?

philippeauriach:
Hi vaiton !

- for now logs are sent to a hosted instance with an open source plan (the same one where we have the robotoff servers error logs). We plan to move to a self hosted instance when possible. 
- there is a similar solution implemented in the playstore, but it is clearer and easier to use in sentry. Also, it lets us have contextual informations to when the crash occurs (list of screen names for example). 
- user privacy is fully respected as no personal information is sent with the log. I also plan to add a « disable crash reporting » settings in the settings of the app.

VaiTon:
> 
> 
> Hi vaiton !
> 
>     * for now logs are sent to a hosted instance with an open source plan (the same one where we have the robotoff servers error logs). We plan to move to a self hosted instance when possible.
> 
>     * there is a similar solution implemented in the playstore, but it is clearer and easier to use in sentry. Also, it lets us have contextual informations to when the crash occurs (list of screen names for example).
> 
>     * user privacy is fully respected as no personal information is sent with the log. I also plan to add a « disable crash reporting » settings in the settings of the app.

I would like to go opt-in, since we don't really have a user policy. @teolemon @stephanegigandet I would like your opinion.

teolemon:
I'm merging.

hangy:
> I would like to go opt-in, since we don't really have a user policy.

Would it be useful to list this hosted service in a GDRP policy, so that users can file a GDPR request?

sentry.properties
```diff
@@ -0,0 +1,4 @@
+defaults.project=openfoodfacts-android
+defaults.org=openfoodfacts
+auth.token=4664a0f9a08e4b2facef49f778b82321993e7e8b0783439fa81f7741b36249b3
```
hangy, line 3: @philippeauriach Does this token need to be public in Git(Hub)? According to https://github.com/getsentry/sentry-react-native/issues/112#issuecomment-309980345, it's only used to upload symbols to Sentry. Therefore, it's only required for releases. With the token in Git, everyone would be able to publish new symbols (potentially overriding existing version?) to Sentry.

philippeauriach:
For transparency yes we could include this service in a gdpr list, but only to specify that we do not send any personal information to it. In case of a gdpr request such as « list all my data », there won’t be anything from sentry.

sentry.properties
philippeauriach: It is not mandatory to make it public of course, it is only used during assemble to upload symbols as you mentioned. It is possible to use an environment variable : `export SENTRY_AUTH_TOKEN=your-auth-token` . Making the change

hangy: I was just wondering if this might be a tiny security issue. 🙂

philippeauriach: You're right, I updated the code, and invalidated the commited token. A new one has been generated and is set on @teolemon computer. I updated the ios sentry branch too

PR #3290, opened 2020-06-21
SonarLint

hangy:
Fixes a number of issues raised by SonarLint in IntellIJ.


app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AddProductNutritionFactsData.java
```diff
@@ -88,6 +88,9 @@
         PREFIX_NUTRIMENT_LONG_NAME + "vitamin-k",
         PREFIX_NUTRIMENT_LONG_NAME + "zinc"));
 
+    private AddProductNutritionFactsData() {
```
VaiTon, line 91: I'm not sure about this one. Are you sure we don't need to initialize this fragment anywhere else?

app/src/main/java/openfoodfacts/github/scrachx/openfood/models/BottomScreenCommon.java
```diff
@@ -8,6 +8,9 @@
 import org.json.JSONObject;
 
 public class BottomScreenCommon {
```
VaiTon, line 10: We could move this out of /models/

app/src/main/java/openfoodfacts/github/scrachx/openfood/fragments/AddProductNutritionFactsData.java
hangy: I think this might be a misplaced file. It's not actually a `Fragment`, but a utility class that's used by the `AddProductNutritionFactsFragment` class.

app/src/main/java/openfoodfacts/github/scrachx/openfood/models/BottomScreenCommon.java
hangy: Do you have a suggestion? This is the first time I actually had more than a quick glance at this project, so that I'm not quite sure about the app's architecture. I'd probably add this to some kind of ViewModel or some Service.

VaiTon: Move it to utils package. 👍

PR #3297, opened 2020-06-23
Automatic release

philippeauriach:
Fastlane configuration with two lanes : 
- `release` lane: fetch current version name and codes from production, increment, assemble, upload to internal track on gplay, create sentry release.
- `daily` lane: finalize a sentry release with the current version in production. Should be run automatically every night by a CI server. Can also be run manually after a version has been deployed to the production track on gplay.

Because of the need for plugins and dependencies, fastlane is used with bundle:
- first time you checkout, run `bundle install` at the root of the project
- then launch lanes using `bundle exec fastlane release` (for the `release` lane)

The lanes won't work if you do not pass a path to an authorized json key file (see `Appfile`) in the env variable `GOOGLE_PLAY_JSON_PATH`. Package name can also be set using the env variable `PACKAGE_NAME`, but defaults to `openfoodfacts.github.scrachx.openfood` if not set. This would be handy when automating other flavors release (obf, etc)


VaiTon:
> Because of the need for plugins and dependencies, fastlane is used with bundle:
> - first time you checkout, run `bundle install` at the root of the project
> - then launch lanes using `bundle exec fastlane release` (for the `release` lane)

Could you add this to the README ?

> - `daily` lane: finalize a sentry release with the current version in production. **Should be run automatically every night by a CI server.** Can also be run manually after a version has been deployed to the production track on gplay.

Could you add a github action to execute this every day or night? I could link you some docs, or if you want we can merge and then think about it.

philippeauriach:
Added the readme section.
For the github action, I prefer to wait that someone with access to the play store and sentry (eg @teolemon) has time to test and validate that everything works fine before launching it using actions. Also, this would mean putting the google play json key file and the sentry token somewhere secure for the github action to use it ? I have no experience using Actions but I guess there is a way of having private env variables.

VaiTon:
I also don't understand the difference between daily and release. What do you mean with "production"?

philippeauriach:
No problem, the concept of “release” is a bit weird and different between services. I’ll try to explain:

The « release » lane is building the app and uploading it to the play store in the “internal” track (there is four tracks : internal, alpha, beta, production). The release lane also create a “release” in sentry but dot not finalize it (it tells sentry that we created a new release with a version name and code, but that it is not live yet). 

When teolemon or another admin decides that the app is stable enough, they use the google play console to pass the app from the internal to beta to production track. The daily lane, ran regularly, only check the version actually in the production track, and tell sentry that this version code and name is “finalized”, which in sentry’s words means released to the public. This way, we do not have to think about finalizing in sentry when publishing a release to the production track.

teolemon:
Looks good to me 👍 
thank you very much @philippeauriach

Issue #3180, opened 2020-05-26
 When adding a photo from the picture manager, the app redirects to an empty scan window


teolemon:
**Describe the bug**
 When adding a photo from the picture manager, the app redirects to an empty Scan window. I'd expect to be back to the product, as it was the case before.

**To Reproduce**
scan -> swipe up -> tap picture -> tap chekcbox
resultats :
Playstore : return to product page & scan OK
Master : retour to scan KO

teolemon:
slide up
tap the photo
tap on select new photo button
select new photo
tap on the checkbox to validate
you’re back to the scanner on master
and in prod, you’re back to the product page
(which is the desirable behaviour)

runyud:
Hi I would like try to debug this issue if possible

PR #3300, opened 2020-06-25
When adding a photo from the picture manager, the app redirects to an empty Scan window #3180

runyud:
some ideas on fixing issue https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/3180
The fact that the productimagemanagementactivity started by summaryproductfragment, ingredientproductfragment, and nutrition fragment in FullscreenActivityOpener.java might be related to the fact that these fragments are not added to the backstack. So when the user calls finish in productimagemanagementactivity, the activity ends and redirects to the previous screen (works in history but not in scan).
Some ideas on fixing the issue: add the current fragment to the backstack so when user finishes he can call popBackStack upon exit (still doesn't work properly).

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/FullScreenActivityOpener.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/FullScreenActivityOpener.java
@@ -9,6 +9,8 @@ import android.widget.Toast;
 
 import androidx.core.app.ActivityOptionsCompat;
 import androidx.fragment.app.Fragment;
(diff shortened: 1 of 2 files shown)


VaiTon:
I don't remember if we merged but I made it return to the scan with the product selected, it was just a problem of the bottom bar not being expanded.

teolemon:
We still have the issue, at least when we add an ingredient image from the product page, on a product that does not have an ingredient image yet.

teolemon:
@VaiTon fixed the issue. closing

PR #3306, opened 2020-06-27
Update Appfile

teolemon:


VaiTon:
@teolemon could you remind me the difference between package names?

teolemon:
I added the explanation directly in the README: https://github.com/openfoodfacts/openfoodfacts-androidapp/blob/master/README.md

PR #3313, opened 2020-06-28
document onboarding

teolemon:


VaiTon:
Could you move it as a javadoc of the class?

teolemon:
Is that what you expected ?

PR #3317, opened 2020-07-03
Product silhouette, Open Food Facts logo and Nutri-Score vector assets

teolemon:
Add missing assets (they are not used by the code yet)

1. Vector replacement for placeholder_thumb.webp that's the same as the web placeholder
2. Add Nutri-Score vector assets. I had issues with distorted versions, but those ones should be fine.
3. I added both dark mode and light mode for the Open Food Facts logo. Not sure if the bars of the barcode will render properly.
4. dark mode and light mode for the OBF, OPF and OPFF logos


teolemon:
image

PR #3327, opened 2020-07-10
feature/dependabot-config

teolemon:
feature/dependabot-config


VaiTon:
At the moment dependabot does not support kotlin gradle file. See https://github.com/dependabot/dependabot-core/issues/2238.

PR #3343, opened 2020-07-16
Fix/3332: fields not updated on product editing

### LINKED ISSUE #3332: Impossible to delete categories
**Describe the bug**
Impossible to delete categories
**To Reproduce**
1. Go to 'Edit'
2. Delete categories
3. Click save
4. Categories are still around

VaiTon:
Fixes #3332.


teolemon:
I'm not quite sure why we did that in the first place, but this looks ok to me 👍

VaiTon:
> 
> 
> I'm not quite sure why we did that in the first place, but this looks ok to me 👍
> Are we resending values in the proper way ? language ? tag vs actual value ? Those are the only caveats that could justify it. But I don't see issues on category addition, so I guess that's fine.

I would like to hear @stephanegigandet on the API side.

PR #3347, opened 2020-07-19
Draft changelog for version 342

teolemon:


VaiTon:
FYI you cannot move resources to off if they are not present on main. Off, obf, etc can only override resource as far as I know. 

Fixing. :-)

teolemon:
@VaiTon can we merge it in `develop` rather ?

VaiTon:
> 
> 
> @VaiTon can we merge it in `develop` rather ?

Why? I don't think it's necessary but if you want sure we can.

PR #3348, opened 2020-07-19
Fixes #3341 Add Nutri-Score in products to be completed

### LINKED ISSUE #3341: Add the Nutri-Score in the "Products to be completed" activity by refactoring it
- [ ] Add the Nutri-Score in the "products to be completed" activity so that power users can see which new products have a Nutri-Score, and which ones don't have one
- [ ] Get the occasion to standardize the activity with the other list activities that already show the Nutri-Score

Varsha-29:
The nutri score is populated in the additional properties with "nutrition_grade_tags" as key. But to display the score, only NutritionGradeFr was considered. Modified the method to also check the additional properties of the product to check for the score


teolemon:
Both those fields are the exact same value: https://world.openfoodfacts.org/api/v0/product/5410041001204.json?fields=nutrition_grade_fr,nutrition_grades_tags
It's just some API mess on the server side. I guess the issue is purely a display one.

Varsha-29:
These are the fields which are used in querying the response for incomplete products API : "brands,product_name,image_small_url,quantity,nutrition_grades_tags,code"
Because of this "nutrition_grade_fr" is null in the Product model.

Varsha-29:
@teolemon , could you let me know if anything has to be changed as a part of this PR?

teolemon:
It seems to behave correctly. @VaiTon ok to merge codewise ?

VaiTon:
Checking and then merging.

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
```diff
@@ -355,7 +355,11 @@ public static int getNovaGroupDrawable(@Nullable String novaGroup) {
     }
 
     public static int getSmallImageGrade(Product product) {
-        return getSmallImageGrade(product == null ? null : product.getNutritionGradeFr());
+        String nutritionGrade = null;
+        if (product != null) {
+            nutritionGrade = product.getNutritionGradeFr() != null ? product.getNutritionGradeFr() : product.getNutritionGradeTag();
```
VaiTon, line 360: Why do we prefer the FR grade? Any particular reason? @teolemon @Varsha-29

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
```diff
@@ -355,7 +355,11 @@ public static int getNovaGroupDrawable(@Nullable String novaGroup) {
...
+        }
+        return getSmallImageGrade(nutritionGrade);
```
VaiTon, line 362: Can we inline this return? I prefer having multiple exit points than having to remember the value of `nutritionGrade`.

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
teolemon: That's the name we initially gave to the Nutri-Score when it didn't have a name.

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
Varsha-29: Done

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.java
VaiTon: Ok, but here we prefer using the FR tag. Are the FR tag and the "global" the same? Any differences?

PR #3349, opened 2020-07-19
design: dark mode for nova logos, fix logos

teolemon:
add dark mode NOVA logos
image


teolemon:
@VaiTon good to merge ?

PR #3359, opened 2020-07-23
reapply changes by hangy

teolemon:


teolemon:
2020-07-23T13:48:39.8464876Z /home/runner/work/openfoodfacts-androidapp/openfoodfacts-androidapp/app/src/main/res/values-sv/strings.xml:234:74: Resource and asset merger: The prefix "tools" for attribute "tools:ignore" associated with an element type "string" is not bound.
2020-07-23T13:48:39.8465246Z     org.xml.sax.SAXParseException; lineNumber: 234; columnNumber: 74; The prefix "tools" for attribute "tools:ignore" associated with an element type "string" is not bound.

teolemon:
ok, was missing properties.

PR #3368, opened 2020-07-26
chore(rx) made postImg RxJava like

VaiTon:


teolemon:
Works well on my device.

Issue #2963, opened 2019-12-23
Add onboarding tooltip on first use for image editing


teolemon:
![image](https://user-images.githubusercontent.com/1689815/71366971-d3b56980-25a3-11ea-817c-ab9c31f6916a.png)

Add onboarding tooltip on fist use for image editing

teolemon:
Here's a code pointer for a similar tipbox already in trunk : https://github.com/openfoodfacts/openfoodfacts-androidapp/commit/6bb72d0d0b03d4c8a4ec911a9e0f70be1bdb4c3c

Varsha-29:
I will work on this feature

teolemon:
This is fixed 🎉 
thanks @Varsha-29 and @VaiTon

PR #3403, opened 2020-08-08
Feat : #2963 - Add onboarding tooltip on first use for image editing

Varsha-29:
*Display tool tip for ingredients image
*Load it the first time only when the image is available. The tooltip will not be shown if the user first taps on an ingredient which has no image associated
*Added customization to the tooltip to configure background and text color

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/TipBox.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/views/TipBox.java
@@ -47,7 +47,12 @@ public class TipBox extends LinearLayout {
         int marginStart = attributes.getDimensionPixelSize(R.styleable.TipBox_arrowMarginStart, 0);
         int marginEnd = attributes.getDimensionPixelSize(R.styleable.TipBox_arrowMarginEnd, 0);
         int arrowAlignment = attributes.getInt(R.styleable.TipBox_arrowAlignment, Gravity.START);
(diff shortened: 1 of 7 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/views/TipBox.java
```diff
@@ -59,21 +57,17 @@ public TipBox(Context context, @Nullable AttributeSet attrs) throws Exception {
         });
 
         prefs = PreferenceManager.getDefaultSharedPreferences(getContext());
-        boolean show = prefs.getBoolean(identifier, true);
-        if (!show) {
-            return;
+        //Tooltip should be shown the first time the actual content is available
+        boolean canDisplayImmediately = attributes.getBoolean(R.styleable.TipBox_shouldDisplayImmediately, false);
```
VaiTon, line 61: I would prefer if we got the attributes that we need immediately so we can call `recycle()` as fast as we can. 

Please move `canDisplayImmediately` near the others `attributes.get...` and create a var for `attributes.getColor` s.

Varsha-29: Done

Varsha-29:
@VaiTon , I didn't make any changes to unit tests and the tests have passed locally in my machine. Not sure why it had failed here

VaiTon:
Thank you! 🥂

PR #3489, opened 2020-08-30
Should check_first_time be translatable ?

teolemon:
Should check_first_time be translatable ?


VaiTon:
Nop, it is used for settings. I guess we should not translate that :)

teolemon:
I pushed a fix to values/strings.xml to develop as well, so that crowdin propagates it.

PR #3513, opened 2020-09-06
Implements onRefresh event at SummaryProductFragment

paulocoe:
Implements onRefresh event to allow pull-down refresh when browsing a expanded view of a scanned product.


VaiTon:
It works! But it doesn't query for product state, only for robotoff questions... Strange..

VaiTon:
Thank you! 🥳

paulocoe:
Can I close the issue related to this PR or do we need something else done?

PR #3519, opened 2020-09-11
Fix for cypher Android 7.0

teolemon:
Fix for cypher Android 7.0


VaiTon (changes requested):
The code is outside of a method

PR #3549, opened 2020-10-16
Remove distributionSha256Sum from gradle-wrapper.properties

huzaifaiftikhar:
**Description**
1. We get the below error after cloning and syncing the project, this happens because we had `distributionSha256Sum` present, since the sha256 can be different for different os versions like Window, Darwin or Linux then Android Studio throws an error.
2. We also got a query on slack for the same. The slack thread can be viewed [here](https://openfoodfacts.slack.com/archives/C02KVRT2C/p1602744257092500).
3. Android Studio itself suggests to remove distributionSha256Sum and sync the project.
4. Related issue https://github.com/gradle/gradle/issues/9361

**Screenshot**
image (1)


VaiTon:
Thanks! 👍

PR #3551, opened 2020-10-20
Fix Null Pointer Exception on No product found from deep link

### LINKED ISSUE #3498: Deep link to the app is not working
Deep link is not working. It's also not working for subdomains like fr.off, it.off…

1. Steps: click on a world.off.org link
2. Click on "Open in Open Food Facts"
3. See a black screen like this

image

sarah541:
####  Description
Fix Null Pointer Exception on No product found from deep link.
- The Deep Link issue works. However when a product is not found the app was crashing. That is fixed.
- Also the handling of when a product is not found is incorrect. User should be shown that product is not found. Currently only closes the screen - Can be fixed in a separate issue.

#### Related issues
- Fixes #3498


huzaifaiftikhar (approved):

VaiTon:
Thank you @sarah541! 🎉🎉

Issue #3540, opened 2020-10-10
The language preference is ignored unless it's set system-wide


pataquets:
**Describe the bug**
I'm a Catalan language translator and I've found that setting Catalan as language in preferences is not working, unless your Android system dude language is also Catalan.
Same for Spanish.

**To Reproduce**
1. Set system language to Spanish.
2. OFF correctly defaults to Spanish.
3. Go to preferences and select Catalan.
4. UI language is now in English.

Tested also swapped ES/CA system and lang on another device using CA/ES for sys/lang.

**Smartphone:**
 - Device: Moto G5s
 - OS: Android 7.0
 - Browser: Android app
 - Version: 3.4.8 (latest from Google Play)

PR #3554, opened 2020-10-23
Fix NPE while changing language preference

sarah541:
####  Description

Fix Null Pointer Exception on changing language preference.
Occurs when you change the language preference for the second time.
**LOGS**
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'void io.reactivex.disposables.CompositeDisposable.dispose()' on a null object reference
        at openfoodfacts.github.scrachx.openfood.features.HomeFragment.onDestroy(HomeFragment.java:100)
        at androidx.fragment.app.Fragment.performDestroy(Fragment.java:3205)
        at androidx.fragment.app.FragmentStateManager.destroy(FragmentStateManager.java:758)

#### Related issues
#3540

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/HomeFragment.java
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/HomeFragment.java
@@ -97,7 +97,9 @@ public class HomeFragment extends NavigationBaseFragment {
     @Override
     public void onDestroy() {
         // Stop the call to server to get total product count and tagline
-        compDisp.dispose();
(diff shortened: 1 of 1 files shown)


VaiTon:
@sarah541 right one! Except I would like it more if it was initialized in the field declaration so we don't have to check for nullability.

Something like:
```java
private final CompositeDisposable compDisp = new CompositeDisposable();
```

What do you think about it?

sarah541:
> initialized in the field declaration

@VaiTon Makes sense. I will push a new commit.

PR #3558, opened 2020-10-25
Bugfix: Fix format issue crash

mohsin:
####  Description
Fixes `IllegalFormatConversionException: %d can't format java.lang.String arguments at openfoodfacts.github.scrachx.openfood.features.HomeFragment.setProductCount(HomeFragment.java:225)`.

The txtHomeOnline string should only have %s as in the other locale string. However in these 6 locales it is incorrect and this PR fixes this issue.


teolemon (approved):

teolemon:
Thanks @SaifurRahmanMohsin 👍

Issue #3559, opened 2020-10-26
IllegalStateException : attempting to register while current state is RESUMED. LifecycleOwners must call register before they are STARTED.


sarah541:
**Describe the bug**
startActivityForResult() was depreciated. The replacement for that was recently done in the code but not implemented correctly causing ISE. 
**To Reproduce**
1. Click to menu item **Sign in** from drawer on home screen - will experience a crash.
The same crash is being experienced at different places in the App. Few noticed ones are
a. Search a valid product on home screen .
b. Click on the product that comes in the search result.
c. Go to INGREDIENTS tab.
d. Click on edit. 
    i) If you are logged in - will experience a crash.
    ii) Not logged in - Click on Sign In on the pop up - will experience a crash

**Expected behavior**
Respective screens should open.

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**
**LOG**
java.lang.IllegalStateException: LifecycleOwner openfoodfacts.github.scrachx.openfood.features.MainActivity@91ab37b is attempting to register while current state is RESUMED. LifecycleOwners must call register before they are STARTED.

PR #3560, opened 2020-10-26
Fix crashes caused by IllegalStateException on the 3 screens

sarah541:
####  Description
startActivityForResult() was depreciated. The replacement for that was recently done in the code but not implemented correctly causing ISE.
#### Related issues
#3559


sarah541:
If a crash similar to this issue(Log) is experienced elsewhere in the App, one can make changes similar to this PR or can ask me to fix it.

sarah541:
Forwarded to #3562

Issue #3531, opened 2020-09-30
Add a button to take a picture of the recycling instructions


teolemon:
### Overview
Add a button on the product page to take a picture of the recycling instructions and upload it to the server.

- This can be achieved by cloning the "Take another picture" code. (Most simple implementation)
- This can be achieved by cloning the "Take ingredients picture" code
- This is **very important** for the Environnemental Scoring project.

### Feature visible in the Web version
![image](https://user-images.githubusercontent.com/1689815/95175767-3e930280-07bc-11eb-88a2-924652dce45f.png)

### On Android
You can see a similar button on the ingredient tab here: https://github.com/openfoodfacts/openfoodfacts-androidapp/blob/c30f75f70fd20c84a1d3d3cf803e9b120b93bab6/app/src/main/res/layout/fragment_ingredients_product.xml#L49
We would insert it in the Environnement tab here: https://github.com/openfoodfacts/openfoodfacts-androidapp/blob/c30f75f70fd20c84a1d3d3cf803e9b120b93bab6/app/src/main/res/layout/fragment_environment_product.xml 
and as a result, make the Environnement tab always visible so that people can help with taking photos.
### Flow
It works exactly like the ingredients field, tab and addition workflow. **This means although I've redocumented everything, you can mostly clone the ingredients code, or refactor it to be able to handle the packaging case**
1. Upload a packaging_image
API doc to upload an image: https://documenter.getpostman.com/view/8470508/SVtN3Wzy?version=latest#6159dfb9-dd05-4ef8-8bf0-2c2ea7585760
Same process as ingredients but field is: ` packaging_fr` , ` packaging_en` …
2. Request OCR output
Request: `cgi/packaging.pl?code=3454681006511&id=packaging_fr&process_image=1&ocr_engine=google_cloud_vision`
Response: `{"packaging_text_from_image":"tia ac ","packaging_text_from_image_orig":null,"status":0}`
API to perform OCR on it: https://documenter.getpostman.com/view/8470508/SVtN3Wzy?version=latest#c911e8c8-2b88-44cd-8512-52fa55131396
3. Get the user to validate it 
4. Send the validated result to packaging_text
Same process as ingredients but field is: ` packaging_text_fr` , ` packaging_text_en` …

Server-side issue: openfoodfacts/openfoodfacts-server#4286

charchithere:
I would Like to Work on this Issue

teolemon:
Welcome @charchithere 
[...]

PR #3565, opened 2020-10-29
Feature: To take a picture of the recycling instructions

mohsin:
* Makes the environment tab appear by default.
* Allows users to take photo of the recycling/packaging instructions.

#### Related PRs
Adds to closing of #3531

#### Screenshots
Screenshot_20201029-220339
Screenshot_20201029-220400


teolemon:
It seems that the uploads go on ok, but unlike for ingredients, there's no feedback at all (including when I try to refresh the product)

teolemon:
arg sorry, I just saw that the rebase did not go on ok.

mohsin:
@teolemon yes, that's because the API doesn't work as intended. I raised an issue [on this earlier here](https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/3531#issuecomment-716461002) and am awaiting response and fix on the server side for the same.

This PR actually does look at the packaging field (which the server isn't sending even thou the docs have been updated as per that comment) and already looks for the right field i.e. packaging_en so should the server API start responding with the packaging image, it should start appearing. If it doesn't I would make an additional commit to ensure it works... but only if [that blocker](https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/3531#issuecomment-716461002) is fixed first.

teolemon:
good catch @stephanegigandet ?
https://world.openfoodfacts.org/cgi/product_image_crop.pl?code=8901063017252&id=packaging_en&imgid=14&x1=0&y1=0&x2=0&y2=0&coordinates_image_size=full&angle=0&normalize=false&white_magic=false
{"image":{"display_url":"packaging_en.41.400.jpg"},"status":"status ok","imagefield":"packaging_en"}

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/ImagesManageActivity.java
```diff
@@ -87,7 +87,7 @@
     private static final int REQUEST_UNSELECT_IMAGE_AFTER_LOGIN = 4;
     public static final int REQUEST_EDIT_IMAGE = 1000;
     private static final int REQUEST_CHOOSE_IMAGE = 1001;
-    private static final List<ProductImageField> TYPE_IMAGE = Arrays.asList(ProductImageField.FRONT, ProductImageField.INGREDIENTS, ProductImageField.NUTRITION);
+    private static final List<ProductImageField> TYPE_IMAGE = Arrays.asList(ProductImageField.FRONT, ProductImageField.INGREDIENTS, ProductImageField.NUTRITION, ProductImageField.PACKAGING);
```
VaiTon: I would rename the field to IMAGE_TYPES and make it an array. I don't think we need list operations. 

Also, if possible, I would move the field to the ApiFields class. 

What do you think?

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/ProductViewActivity.java
```diff
@@ -104,11 +104,8 @@ public static ProductFragmentPagerAdapter setupViewPager(@NonNull ViewPager2 vie
 
         if (AppFlavors.isFlavors(AppFlavors.OFF)) {
             adapter.addFragment(FragmentUtils.applyBundle(new NutritionProductFragment(), fBundle), menuTitles[2]);
-            if ((productState.getProduct().getNutriments() != null &&
-                productState.getProduct().getNutriments().contains(Nutriments.CARBON_FOOTPRINT)) ||
-                (productState.getProduct().getEnvironmentInfocard() != null && !productState.getProduct().getEnvironmentInfocard().isEmpty())) {
-                adapter.addFragment(FragmentUtils.applyBundle(new EnvironmentProductFragment(), fBundle), "Environment");
-            }
+            adapter.addFragment(FragmentUtils.applyBundle(new EnvironmentProductFragment(), fBundle), menuTitles[4]);
```
VaiTon, line 107: Good one!

VaiTon (changes requested):
Also, have you written the code to update the db because of a change in the model? I can't seem to find it. Thank you!

teolemon:
@SaifurRahmanMohsin the behaviour should be fixed in this PR that @stephanegigandet is going to deploy to prod soon:
https://github.com/openfoodfacts/openfoodfacts-server/issues/4442

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/ImagesManageActivity.java
mohsin: Resolved by #59f2681

VaiTon:
> 
> 
> Also, have you written the code to update the db because of a change in the model? I can't seem to find it. Thank you!

What about this? @SaifurRahmanMohsin

mohsin:
I think I missed that out. Thanks for letting me know, I'll update it @VaiTon

teolemon:
@stephanegigandet confirmed that `image_packaging_url` will work once it's deployed. Thus https://ssl-api.openfoodfacts.org/api/v0/product/8901063017252.json?fields=image_packaging_url should return the packaging url

teolemon:
@SaifurRahmanMohsin @VaiTon good news, the URL is live.

teolemon:
```
 Process: openfoodfacts.github.scrachx.openfood.debug, PID: 11303
    java.lang.NullPointerException: Attempt to invoke interface method 'java.lang.Object java.util.Map.get(java.lang.Object)' on a null object reference
        at openfoodfacts.github.scrachx.openfood.models.Product.getImagePackagingUrl(Product.java:215)
        at openfoodfacts.github.scrachx.openfood.features.product.view.environment.EnvironmentProductFragment.onViewCreated(EnvironmentProductFragment.java:65)
        at androidx.fragment.app.Fragment.performViewCreated(Fragment.java:2974)
```

SaifurRahmanMohsin:
@teolemon You can ignore that, I had added a commit to use images array instead. I'll revert/reset back one commit and set the app to use image_packaging_url to stay consistent with recycling code as well.

mohsin:
@teolemon I've removed that commit. I had added that to use `images` instead but I've reverted back to using `image_packaging_url ` and you can recheck the code now.

teolemon:
latest commit does not work for me. The button to take the photo does appear. You can take the photo and then nothing happens (the "Take a photo" icon is not replaced, and the image isn't uploaded silently either)

SaifurRahmanMohsin:
I see commits that I had made earlier being re-pushed @VaiTon. Not sure what's going on here.

@teolemon Did you check will all the changes made? Because [the code to open taking the photo](https://github.com/openfoodfacts/openfoodfacts-androidapp/pull/3565/files#diff-cbd9db7e24f1164d4174482fc9b75368622ca9faa359812b72aed87236498c0bR55-R61) is there already so there has to be some reaction for sure.

teolemon:
The photo selector is now working, but photo taking has the same symptom.

mohsin:
@teolemon I've added the code to ensure the photo taking happens and the user can see the photo after it's taken immediately. You can re-confirm this and let me know if it works fine now.

teolemon:
@SaifurRahmanMohsin latest commit works well 👍

teolemon:
@VaiTon @SaifurRahmanMohsin Let's merge since we can probably do the Editing/OCR part in a separate PR

teolemon:
@VaiTon ok to merge this one ?

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/environment/EnvironmentProductFragment.java
```diff
@@ -10,41 +12,102 @@
...
+    /**
+     * boolean to determine if image should be loaded or not
+     **/
+    private boolean isLowBatteryMode = false;
+    private Product product;
+
+    @Override
+    public void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        disp = new CompositeDisposable();
```
VaiTon: I would inizialize this in the declaration and make it final.

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/environment/EnvironmentProductFragment.java
```diff
@@ -10,41 +12,102 @@
...
+    }
 
     @Override
     public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
+        api = new OpenFoodAPIClient(requireActivity());
```
VaiTon: Put it in onCreate

VaiTon (approved):

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/environment/EnvironmentProductFragment.java
SaifurRahmanMohsin: @VaiTon Is there any specific reason why this OpenFoodAPIClient is initialized in the fragments each time rather than being injected as a singleton (I see dagger library is already present in the app).

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/environment/EnvironmentProductFragment.java
mohsin: Cool, done.

Wouldn't it be better to use Uber's [AutoDispose](https://uber.github.io/AutoDispose) and get rid of all the CompositeDisposables throughout the app?

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/environment/EnvironmentProductFragment.java
mohsin: I've made the change @VaiTon

Also, is there any specific reason why this OpenFoodAPIClient is initialized in the fragments each time rather than being injected (I see dagger library is already present in the app) like:
```
@Inject
OpenFoodAPIClient api;
```

teolemon:
congratulations @SaifurRahmanMohsin 👍 and thank you @VaiTon for the review.

mohsin:
@teolemon After going through the code, I can think of at-least a dozen improvements and changes which I'm listing here:

* Why not use [Timber](https://github.com/JakeWharton/timber) instead of Log? You are essentially taking the class name and doing the same as what Timber does so you can simplify this by using Timber everywhere? Also, using Log message directly is literally bad because when one builds in production, these logs pollute the global device logs and even slow down the device whereas with timber you can specify it to work debug-only.
* Is there any specific reason why OpenFoodAPIClient/WikiApiClient are initialized in the fragments each time rather than being injected (I see dagger library is already present in the app) like:
```
@Inject
OpenFoodAPIClient api;
```
It maybe good to create a Dagger provider singleton for these and use that instead.
* The GreenDao db doesn't have support for RxJava so you're missing out on it. Since this project uses RxJava it might be useful to use something like Room which has an adapter to link it with RxJava--you could even make the offline support work using RxJava operator.
* Regarding offline support, even thou there is code for it, I don't think there is sufficient offline support now. Switching to Room also helps here.
* There are compositedisposables many of the fragments as it should be generally when using RxJava. But these can be removed in favor of [Uber's Autodispose](https://uber.github.io/AutoDispose) which simplifies many things.
* Why not use Jetpack's new NavigationUI? It's much cleaner and saves code.
* It maybe a good idea to switch from Dagger to Hilt (built on top of dagger) that has much cleaner code.
* It might be a good idea to switch to Kotlin. This is extensive work but worthwhile.

These are my suggestions based on what I've seen. I'd be able to even contribute and make some of these changes if you can decide if these are worthwhile changes and we should. If not, I understand.

teolemon:
@SaifurRahmanMohsin those are many interesting points. Can you put them in a new issue so that we can discuss them ?
Would you join #android in the Open Food Facts slack (https://slack.openfoodfacts.org) ?

PR #3569, opened 2020-11-03
feature: ecoscore implementation and display

### LINKED ISSUE #3529: Add support for displaying the Eco-Score
### Add support for displaying the Eco-Score. 

- [ ] Clone the system that displays Nutri-Score on the scan card renaming it 'Eco-Score'
- [ ] Change the server request url we do on scan to include the ecoscore field 
- [ ] If (and only if) the server sends back a proper value (a, b, c, d or e), hide the classic CO2 value and display the new score
`https://world.openfoodfacts.org/api/v0/product/3414280980209.json?fields=ecoscore` (Here it shows null)
- [ ] Please use the placeholders visuals already in the app as placeholders until we have the visuals. It would be better to download them on the fly so that we can remote update them.
https://static.openfoodfacts.org/images/icons/ecoscore-a.svg
https://static.openfoodfacts.org/images/icons/ecoscore-b.svg
https://static.openfoodfacts.org/images/icons/ecoscore-c.svg
https://static.openfoodfacts.org/images/icons/ecoscore-d.svg
https://static.openfoodfacts.org/images/icons/ecoscore-e.svg

### Places where it should be shown

The same actions have to be done on the summary tab, and on top of the `Environment` tab (where we also display the score)

- [ ] Scan card
- [ ] `Environment` tab
- [ ] Summary tab if attributes are not implemented yet

(server-side issue: https://github.com/openfoodfacts/openfoodfacts-server/issues/4262)

VaiTon:
Implementation of #3529

### Features implemented
- [x] Change the server request url we do on scan to include the ecoscore field
- [x] 'Eco-Score' on the scan card
- [x] If the server sends back a proper value, hide the classic CO2 value and display the new score
- [ ] Download icons on the fly so that we can remote update them.

#### Places where it should be shown
- [x] Scan card
- [ ] Environment tab (waiting for #3565)
- [x] Summary tab if attributes are not implemented yet (spoiler: they are not)


app/src/main/java/openfoodfacts/github/scrachx/openfood/network/ApiFields.java
```diff
@@ -87,6 +87,7 @@ private Defaults() {
         public static final String ADD_BRANDS = "add_brands";
         public static final String BRANDS = "brands";
         public static final String LC = "lc";
+        public static final String ECOSCORE = "ecoscore";
```
teolemon: ecoscore to ecoscore_grade

app/src/main/res/values/strings.xml
```diff
@@ -354,6 +354,7 @@
         <item>nutrition_grade_fr</item>
         <item>nutrition_grades_tags</item>
         <item>nutrient_levels</item>
+        <item>ecoscore</item>
```
teolemon: ecoscore to ecoscore_grade

teolemon:
works great. good to merge.

teolemon (approved):

PR #3597, opened 2020-11-17
Fastlane beta script now uses the version indicated in the branch name

philippeauriach:
For example, if pushing on the release/3.4.1 branch, then the internal build will use the version 3.4.1.

PR Description
Proposed changes

before, the "fastlane release" script would push a build on the play store internal using the latest version used in the play store. 
This leads to problems when code is pushed on two different "release/xxx" branches (as a push on release/4.3 would trigger a build with version 4.3.1 if any already exists).
Now, the script use the version specified in the name branch, so a push on release/4.3 will trigger a build with the version set to 4.3

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -16,11 +16,10 @@ lane :release do
     PACKAGE_NAME = CredentialsManager::AppfileConfig.try_fetch_value(:package_name)
     gradle_file = "./app/build.gradle.kts"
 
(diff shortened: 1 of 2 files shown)


teolemon (approved):

VaiTon:
@philippeauriach could we also use `release/x.x` ?

philippeauriach:
> @philippeauriach could we also use `release/x.x` ?

Not if we follow the SemVer proposal, a version must always be in the form of x.y.z (but z can be 0).

PR #3645, opened 2020-12-29
Migrate Crowdin sync to GitHub actions

teolemon:
Migrate Crowdin sync to GitHub actions
Hopefully, this will 

1. remove the need for manual work of switching between repositories (it's triggered at each commit in develop, vs manually when I select the android repo as the current synced repository)
2. This means one commit vs 147 commits. So less notification spam :-)

Points to check:
        # Skip untranslated strings in exported files (does not work with .docx, .html, .md and other document files)
        skip_untranslated_strings: true
        # Omit downloading not fully downloaded files
        skip_untranslated_files: false

I believe we can skip untranslated_strings since it will fallback to English. That was a request by Stéphane for another repo. But is that a good idea for Android ? @VaiTon


VaiTon:
@teolemon what should this do?

teolemon:
This switches the system we use for Crowdin. 

- This means one commit vs 147 commits. So less notification spam :-) 
- It also means I won't have to synchronize each repo manually.

.github/workflows/crowdin.yml
```diff
@@ -0,0 +1,79 @@
+name: Crowdin Action
+
+on:
+  push:
+    branches: [ master ]
```
VaiTon:
```suggestion
    branches: [ develop ]
```

.github/workflows/crowdin.yml
```diff
@@ -0,0 +1,79 @@
...
+        # Omit downloading not fully downloaded files
+        skip_untranslated_files: false
+        # Include approved translations only in exported files. If not combined with --skip-untranslated-strings option, strings without approval are fulfilled with the source language
+        # export_only_approved: # default is false
+        # Download translations with pushing to branch
+        # push_translations: # default is true
+        # Commit message for download translations
+        commit_message: "New Crowdin translations by Github Action"
+        # To download translations to the specified version branch
+        localization_branch_name: l10n_master_gh
```
VaiTon:
```suggestion
        localization_branch_name: l10n_develop_gh
```

VaiTon (changes requested):

.github/workflows/crowdin.yml
```diff
@@ -0,0 +1,79 @@
...
+        # upload_language: # optional
+        # Automatically approves uploaded translations
+        # auto_approve_imported: # default is false
+        # Defines whether to add translation if it is equal to source string in Crowdin project
+        # import_eq_suggestions: # default is false
+        # Make pull request of Crowdin translations
+        # Use this option to download translations for a single specified language
+        # download_language: # optional
+        # Skip untranslated strings in exported files (does not work with .docx, .html, .md and other document files)
+        skip_untranslated_strings: true
```
teolemon:
```suggestion
        skip_untranslated_strings: false
```

teolemon: @VaiTon would like your input on that one

VaiTon: Let it be false, let's see its behavior

VaiTon (approved):

PR #3664, opened 2021-01-03
added helper msg for login in flavour except OFF

### LINKED ISSUE #3209: When using OBF, OPF or OPFF, mention you can login with an OFF account
When using OBF, OPF or OPFF, there's no way to know that you can login with an OFF account.

if flavour = obf or opf or opff, we could add: "If you have an Open Food Facts account, you can login to Open Beauty Facts with it"

kartikaysharma01:
####  Description

Added helper text on login screen for OBF, OPF, OPFF

#### Related issues

Fixes #3209

#### Screenshots

In OBF,
image

In OFF,
image

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/LoginActivity.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/LoginActivity.kt
@@ -33,7 +33,8 @@ import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.rxkotlin.addTo
 import okhttp3.ResponseBody
-import openfoodfacts.github.scrachx.openfood.AppFlavors
+import openfoodfacts.github.scrachx.openfood.AppFlavors.OFF
+import openfoodfacts.github.scrachx.openfood.AppFlavors.isFlavors
 import openfoodfacts.github.scrachx.openfood.BuildConfig
 import openfoodfacts.github.scrachx.openfood.R
 import openfoodfacts.github.scrachx.openfood.customtabs.CustomTabActivityHelper
@@ -161,7 +162,7 @@ class LoginActivity : BaseActivity() {
         setContentView(binding.root)
 
         // check flavour and show helper text for user account
-        if (!AppFlavors.isFlavors(AppFlavors.OFF)) {
+        if (!isFlavors(OFF)) {
             binding.txtLoginHelper.setText(R.string.txtloginHelper)
             binding.txtLoginHelper.visibility = View.VISIBLE
         }
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/LoginActivity.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/LoginActivity.kt
@@ -22,6 +22,7 @@ import android.net.Uri
 import android.os.Bundle
 import android.util.Log
 import android.view.MenuItem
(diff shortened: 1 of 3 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/LoginActivity.kt
```diff
@@ -158,6 +160,12 @@ class LoginActivity : BaseActivity() {
         _binding = ActivityLoginBinding.inflate(layoutInflater)
         setContentView(binding.root)
 
+        // check flavour and show helper text for user account
+        if (!AppFlavors.isFlavors(AppFlavors.OFF)) {
```
VaiTon: Could you import `isFlavors` and `AppFlavors.OFF`?

app/src/main/res/layout/activity_login.xml
```diff
@@ -117,6 +117,18 @@
...
 
+            <TextView
+                android:id="@+id/txt_login_helper"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_gravity="center"
+                android:gravity="center"
+                android:paddingTop="@dimen/padding_short"
+                android:paddingBottom="4dp"
+                android:textColor="@color/green_500"
```
VaiTon: I would change the color to grey, as this is not a success message.

VaiTon (approved):

app/src/main/res/layout/activity_login.xml
kartikaysharma01: Do you want me to do that?
Also a suggestion, maybe red color can be used as this is a information we would like to highlight before a user tries to log in.

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/LoginActivity.kt
kartikaysharma01: Can you elaborate. I didn't get you.

VaiTon: Could you add as a static import the method `isFlavors`, and do the same with `AppFlavors.OFF`?

app/src/main/res/layout/activity_login.xml
VaiTon: @teolemon what do you think?

teolemon: red and green convey strong meaning. Green rather than Red in the sense that the user hasn't done anything wrong. Grey is an option as well, but it might not be as visible ?

Anyway, we don't need to overthink it too much in terms of priorities right now (it's going to be on OBF, OPF and OPF vs OFF)

teolemon: so merging as is or after changing to grey are both fine to me

kartikaysharma01: I will change it to grey.

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/LoginActivity.kt
kartikaysharma01: Done.

app/src/main/res/layout/activity_login.xml
kartikaysharma01: I have made the changes.

kartikaysharma01: @VaiTon

VaiTon (approved):

PR #3709, opened 2021-01-09
Rename and add scores

teolemon:
I have no idea if that's enough.

- [ ] There are some queries that I was sure to have field limited (like the /state/products-to-be-completed one)
- [ ] This peculiar one is not showing all the scores, so the truth may be in some other file


VaiTon:
Superseeded by 3d33169ff1d6282bb9f7e717e09115d3ac8b70b7

PR #3766, opened 2021-01-22
Fixed Permission bug

### LINKED ISSUE #3762: Allowing permission but a dialog box appears asking to activate it
**Describe the bug**
Inspite of allowing the permission during runtime, a dialog box appears to activate the permission from settings. (Permission is actually)

**To Reproduce**
1. Open any product
2. Click on _Take extra photos of packaging_ in Summary Tab.
3. Allow the permission.

**Expected behavior**
The camera must open.

mrudultora:
####  Description
Fixed the dialog box to appear only when permission is denied. Or else, opens the camera.

#### Related issues
Fixes #3762 

#### Related PRs
None

#### Screenshots


app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.kt
```diff
@@ -310,7 +310,7 @@ object Utils {
 
 }
 
-fun isAllGranted(grantResults: Map<String?, Boolean?>) = grantResults.containsValue(false)
+fun isAllGranted(grantResults: Map<String?, Boolean?>) = grantResults.containsValue(true)
```
VaiTon: We need to check for every result.
You should use:
```suggestion
fun isAllGranted(grantResults: Map<String?, Boolean?>) = !grantResults.containsValue(false)
```

VaiTon (changes requested):

VaiTon (approved):

VaiTon:
Thank you! 🎉

PR #3775, opened 2021-01-23
Show nutrition table and nutrition per portion button if nutritional values are available

### LINKED ISSUE #2960: Do not show nutrition table and nutrition per portion button if nutritional values are empty
image

Do not show nutrition table and nutrition per portion button if nutritional values are empty

mrudultora:
####  Description
If the size of `nutrimentListItems` is greater than 1 then only show them. By default set there visibility to gone.

#### Related issues
Fixes #2960 

#### Related PRs
None

#### Screenshots


mrudultora:
@VaiTon Here, also I am unable to understand why apk generation is failing. I am able to build the app. Also `assembleDebug` shows no errors.

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/nutrition/NutritionProductFragment.kt
```diff
@@ -324,7 +324,13 @@ class NutritionProductFragment : BaseFragment(), CustomTabActivityHelper.Connect
             nutrimentListItems.add(BoldNutrimentListItem(getString(R.string.nutrition_minerals)))
             nutrimentListItems.addAll(getNutrimentItems(nutriments, Nutriments.MINERALS_MAP))
         }
-        binding.nutrimentsRecyclerView.adapter = NutrimentsGridAdapter(nutrimentListItems)
+
+        // Show nutrition table and nutrition per portion button if nutritional values are available
+        if (nutrimentListItems.size > 1) {
```
VaiTon, line 329:
```suggestion
        if (nutrimentListItems.isNotEmpty()) {
```

VaiTon (approved):

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/nutrition/NutritionProductFragment.kt
mrudultora: @VaiTon Here, the size of `nutrimentListItems` is 1 even if no values are present. So, it will never be empty. Therefore, the changes you are asking are not valid. There must be at least 2 elements in the arrayList, then only we must set adapter to recyclerView.

PR #3747, opened 2021-01-16
Hided ecoscore in OBF scan

mrudultora:
####  Description
Ecoscore will be visible in **"off"** only .

#### Related issues
Fixes #3746 

#### Related PRs
None

#### Screenshots
img

PR #3796, opened 2021-02-01
Fixed the visibility of the export, trash, sort buttons in History

### LINKED ISSUE #3076: History export does not seem to be working
History export does not seem to be working. A transient tooltip says history is being exported. 

1. No confirmation that the export has completed
2. No notification to open the file (like in the list export feature)
3. No way to share the file using the Android Share Sheet.
4. Possibly a broken export.

We should try to fix or implement as much of 1, 2, 3 and 4 as possible.

### LINKED ISSUE #3795: History remove and export buttons are invisible
**Describe the bug**
History remove and export buttons are invisible

**To Reproduce**
1. Go to History

**Expected behavior**
The buttons should be visible

kartikaysharma01:
####  Description
Fixed a minor issue to fix build fail caused by #3747 
Added check on visibility of history menu buttons

#### Related issues

Fixes #3795 , 
Fixes #3076 

#### Screenshots

image
image

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/scanhistory/ScanHistoryActivity.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/scanhistory/ScanHistoryActivity.kt
@@ -56,6 +56,11 @@ class ScanHistoryActivity : BaseActivity(), SwipeController.Actions {
      */
     private val isLowBatteryMode by lazy { this.isDisableImageLoad() && this.isBatteryLevelLow() }
 
+    /**
+     * boolean to determine if menu buttons should be visible or not
+     */
+    private var showMenuButtons = false
+
     private lateinit var adapter: ScanHistoryAdapter
 
     private var sortType = NONE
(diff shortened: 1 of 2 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/scanhistory/ScanHistoryActivity.kt
```diff
@@ -340,9 +348,10 @@ class ScanHistoryActivity : BaseActivity(), SwipeController.Actions {
                         binding.historyProgressbar.visibility = View.GONE
                         binding.emptyHistoryInfo.visibility = View.VISIBLE
                         binding.scanFirst.visibility = View.VISIBLE
-                        invalidateOptionsMenu()
                         return@flatMapCompletable Completable.complete()
                     }
+                    showMenuButtons = true
```
VaiTon, line 353: Could we remove this variable, keep the check inside onCreateOptionsMenu and just invalidate in on change?

kartikaysharma01: The problem is, initially every time when the view is created, the adapter is empty . So it is not possible to keep the check in the onCreateOptionsMenu, I think we need a variable for that.

VaiTon (approved):

PR #3835, opened 2021-02-16
fix: don't replace fragment on another fragment if they are the same

### LINKED ISSUE #3465: The app seems to duplicate queries
The app seems to duplicate queries
```
XXX.92 - - [23/Aug/2020:11:40:03 +0200] "GET /1.json?fields=null HTTP/2.0" 200 98979 "-" "Open Food Facts Official Android App 3.4.2"
XXX.92 - - [23/Aug/2020:11:40:03 +0200] "GET /files/tagline/tagline-off.json HTTP/2.0" 200 2164 "-" "Open Food Facts Official Android App 3.4.2"
XXX.92 - - [23/Aug/2020:11:40:03 +0200] "GET /files/tagline/tagline-off.json HTTP/2.0" 200 2164 "-" "Open Food Facts Official Android App 3.4.2"
XXX.92 - - [23/Aug/2020:11:40:03 +0200] "GET /1.json?fields=null HTTP/2.0" 200 98979 "-" "Open Food Facts Official Android App 3.4.2"
```

naivekook:
####  Description
when user clicks on "Home" button twice we did a replacement of current fragment on new instance of this fragment and do extra calls to the server

#### Related issues
Fixes #3465

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/MainActivity.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/MainActivity.kt
@@ -168,7 +168,7 @@ class MainActivity : BaseActivity(), NavigationDrawerListener {
         setLanguageInPrefs(this, getLanguage(this))
         setSupportActionBar(binding.toolbarInclude.toolbar)
         supportActionBar?.setDisplayHomeAsUpEnabled(false)
-        swapToHomeFragment()
+        swapToFragment(HomeFragment.newInstance(), HomeFragment.TAG)
 
         // chrome custom tab init
         customTabActivityHelper = CustomTabActivityHelper()
@@ -265,11 +265,11 @@ class MainActivity : BaseActivity(), NavigationDrawerListener {
                 )
                 .withOnDrawerItemClickListener(object : Drawer.OnDrawerItemClickListener {
                     override fun onItemClick(view: View?, position: Int, drawerItem: IDrawerItem<*>): Boolean {
-                        var newFragment: Fragment? = null
+                        var newFragment: Pair<Fragment, String>? = null
                         when (drawerItem.identifier.toInt()) {
-                            ITEM_HOME -> newFragment = HomeFragment.newInstance()
+                            ITEM_HOME -> newFragment = HomeFragment.newInstance() to HomeFragment.TAG
                             ITEM_SEARCH_BY_CODE -> {
-                                newFragment = SearchByCodeFragment()
+                                newFragment = SearchByCodeFragment() to SearchByCodeFragment.TAG
                                 binding.bottomNavigationInclude.bottomNavigation.selectNavigationItem(0)
                             }
                             ITEM_CATEGORIES -> CategoryActivity.start(this@MainActivity)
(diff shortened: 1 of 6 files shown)


naivekook:
@VaiTon @teolemon please take a look

VaiTon:
I don't really like this implementation using tag string. Can't we compare using `is`, for example? Is there a way to get the actual displayed fragment?

naivekook:
> I don't really like this implementation using tag string.

Can I ask you why?

fragment manager doesn't have convenient mechanism to retrieve current fragment, by combining `findFragmentByTag` and `fragment.isVisible` we can understand this

looks like we use `replace` everywhere without backstack so we have only one fragment attached, in this case I can do next thing and it will work too
```
with((currentActivity as FragmentActivity).supportFragmentManager) {
                        val fragment = fragments.lastOrNull()
                        if (fragment == null || fragment !is HomeFragment) {
                            commit {
                                replace(R.id.fragment_container, HomeFragment.newInstance(), HomeFragment.TAG)
                                addToBackStack(null)
                            }
                        }
                    }
```

VaiTon:
> 
> 
> > I don't really like this implementation using tag string.
> 
> Can I ask you why?

Just because it's a useless var if we can check the type as you showed IMHO. 

About your second solution, if you could test it and see if it works it would be awesome!

naivekook:
@VaiTon hey check updated version please
I've been tested it on emulator and it works

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/MainActivity.kt
```diff
@@ -422,9 +417,17 @@ class MainActivity : BaseActivity(), NavigationDrawerListener {
         handleIntent(intent)
     }
 
-    private fun swapToHomeFragment() {
-        supportFragmentManager.addOnBackStackChangedListener {}
-        supportFragmentManager.commit { replace(R.id.fragment_container, HomeFragment()) }
+    private fun swapToFragment(fragment: Fragment) {
+        with(supportFragmentManager) {
```
VaiTon: Really hate saying this but i think `let` would be more clear when calling methods. Could you replace `with` with `let`?

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/MainActivity.kt
```diff
@@ -422,9 +417,17 @@ class MainActivity : BaseActivity(), NavigationDrawerListener {
...
+            val currentFragment = fragments.lastOrNull()
+            if (currentFragment == null || currentFragment::class.java != fragment::class.java) {
```
VaiTon: Why don't we use `currentFragment !is fragment::class`?

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/searchbycode/SearchByCodeFragment.kt
```diff
@@ -76,10 +76,10 @@ class SearchByCodeFragment : NavigationBaseFragment() {
 
     companion object {
         const val INTENT_KEY_BARCODE = "barcode"
-        fun newInstance(barcode: String) = SearchByCodeFragment().apply {
+        fun newInstance(barcode: String? = null) = SearchByCodeFragment().apply {
```
VaiTon, line 79: Why this? We cannot search by code without any code.

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/listeners/CommonBottomListener.kt
```diff
@@ -34,9 +34,14 @@ class CommonBottomListener internal constructor(private val currentActivity: Act
             }
             R.id.home_page, R.id.home -> {
                 if (isCurrentActivity(WelcomeActivity::class.java) || isCurrentActivity(MainActivity::class.java)) {
-                    (currentActivity as FragmentActivity).supportFragmentManager.commit {
-                        replace(R.id.fragment_container, HomeFragment())
-                        addToBackStack(null)
+                    with((currentActivity as FragmentActivity).supportFragmentManager) {
```
VaiTon: Could we extract a method out of this?

VaiTon (changes requested):

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/searchbycode/SearchByCodeFragment.kt
naivekook: sorry I don't get you

actually I see that we dont have any usages of this method so we can remove it safely I guess

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/MainActivity.kt
naivekook: your variant isn't working

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/MainActivity.kt
naivekook: well this is more "style dispute", removed `with`

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/listeners/CommonBottomListener.kt
naivekook: sure, done

naivekook:
ptal @VaiTon

VaiTon (approved):

VaiTon:
Thanks! 🎉

PR #3840, opened 2021-02-20
feat: Add history refreshing (fixes 3162)

### LINKED ISSUE #3162: Make it possible to refresh the history (causes big Nutri-Score problem)
Make it possible to refresh the history on pull down.
A number of values are currently stored in the History, and not refreshed, including 
- the product name
- image
- Nutri-Score. 
We got reports of users complaining the Nutri-Score was not the right one as a result, that the photos were not up-to-date…
We need to set up a refresh mechanism (on pull to refresh ?) so that we can have an up-to-date history. This includes altering the entries, and fetching the new values using the API

naivekook:
####  Description
Added history refreshing via API call. Refactored screen a bit to split up the logic and ui and make the screen more easy to read. The PR is quite big, but I don't see possibility to break it in small parts, sorry for that.

#### Related issues
Fix #3162 

#### Related PRs
none

#### Screenshots
https://drive.google.com/file/d/1G8M0R7RAYfsF1MNATzuwC0GFgm29B4vr/view?usp=sharing

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -92,6 +92,11 @@ dependencies {
     implementation("io.reactivex.rxjava2:rxkotlin:2.4.0")
     implementation("io.reactivex.rxjava2:rxjava:2.2.20")
     implementation("io.reactivex.rxjava2:rxandroid:2.1.1")
+    implementation("com.jakewharton.rxrelay2:rxrelay:2.1.1")
+
+    //Rx optional
+    implementation("com.gojuno.koptional:koptional:1.7.0")
+    implementation("com.gojuno.koptional:koptional-rxjava2-extensions:1.7.0")
 
     //Networking
     implementation("com.squareup.retrofit2:retrofit:2.6.4")
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/productlist/ProductListActivity.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/productlist/ProductListActivity.kt
@@ -384,17 +384,6 @@ class ProductListActivity : BaseActivity(), SwipeController.Actions {
             return notificationManager
         }
 
-        fun getProductBrandsQuantityDetails(brands: String?, quantity: String?): String {
-            val builder = StringBuilder()
-            if (!brands.isNullOrEmpty()) {
-                builder.append(brands.split(",")[0].trim { it <= ' ' }.capitalize(Locale.ROOT))
-            }
-            if (!quantity.isNullOrEmpty()) {
-                builder.append(" - ").append(quantity)
-            }
-            return builder.toString()
-        }
-
         const val KEY_LIST_ID = "listId"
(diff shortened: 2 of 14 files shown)


naivekook:
@VaiTon @teolemon ptal

VaiTon:
I'm really sorry @naivekook but we need to break this down as you're adding tree dependencies and possibly breaking a lot of code. Could we start from refactoring the old code and then adding the new deps and only then adding the new feature. I know it's a lot of work but this pr doesn't only add the feature.

naivekook:
Yeah it's ok, it was to be expected that my PR will be declined

I thought how I can split my PR in a few small PRs, but the problem is that my refactor requires to include these new libraries (by the way these libraries are very small and very stable and tested in production many times). To remove the new functionality from refactoring you just need to remove lines 43-71 from `ScanHistoryViewModel.kt`.

The reason why I made this PR is I just want to show and discuss how we can refactor our screen with view model approach (looking through the code looks like this approach is more easy to implement here). I believe that we have to improve the codebase and write some tests if the project wants to move forward, because, to be honest, my eyes bleeding sometimes looking to it and our latest news about "Bad behaviour" say the same thing.

I have an idea @VaiTon, what if we can hop on call in OFF weekly meeting on Thursday and I will follow you (or someone else) through my code to describe how it works and what I've done?

P.S. sorry if I sounds too linear or offensive, english is not my native, I appreciate the work you've been done guys, just want to make the project better :)

VaiTon:
> 
> 
> Yeah it's ok, it was to be expected that my PR will be declined
> 
> I thought how I can split my PR in a few small PRs, but the problem is that my refactor requires to include these new libraries (by the way these libraries are very small and very stable and tested in production many times). To remove the new functionality from refactoring you just need to remove lines 43-71 from `ScanHistoryViewModel.kt`.
> 
> The reason why I made this PR is I just want to show and discuss how we can refactor our screen with view model approach (looking through the code looks like this approach is more easy to implement here). I believe that we have to improve the codebase and write some tests if the project wants to move forward, because, to be honest, my eyes bleeding sometimes looking to it and our latest news about "Bad behaviour" say the same thing.
> 
> I have an idea @VaiTon, what if we can hop on call in OFF weekly meeting on Thursday and I will follow you (or someone else) through my code to describe how it works and what I've done?
> 
> P.S. sorry if I sounds too linear or offensive, english is not my native, I appreciate the work you've been done guys, just want to make the project better :)

You don't sound offensive at all don't worry! Let's coordinate on slack then

VaiTon:
@naivekook rebase?

naivekook:
Yes just rebased

naivekook:
@VaiTon @teolemon any update about this PR?

teolemon:
small conflict resolution necessary

VaiTon:
@naivekook let's cook this in develop

naivekook:
Sorry I didn't test it well after last rebasing

Issue #2870, opened 2019-08-27
Add an in-app changelog


teolemon:
- Add an in-app changelog that's triggered when opening the app after an upgrade (or series of upgrades) to let the user know what's new from the last time he/she used the app
- Proposed library to use: https://github.com/TonnyL/WhatsNew
- We can either hardcode it, or we can have a JSON file on the server, or somewhere in the code like @VaiTon suggested

hmuhammadazeem:
Hey, I'm interested to work on this. Can I get some details on it?

VaiTon:
@hmuhammadazeem we could create a file CHANGELOG.md and then display it after every update in-app or maybe fetch the details from github.

VaiTon:
Oh my bad, the file CHANGELOG already exists. The second part of my sentence is still valid though.

rprager18:
Hello. I am new to open source and wanted to start my first project. Do you guys think I can take a crack at this

VaiTon:
@rprager18 yes you are very welcome to do so! 

@teolemon recommends this library (https://github.com/TonnyL/WhatsNew). Take a look!

naivekook:
@teolemon @VaiTon Hi!
Is it still TBD? Can I take it?

also do we have another source of changelog?
In current case we need to parse `CHANGELOG.md` via gradle task on each build to get the new changes and store them in application and it quite tricky because this file doesn't have stable formatting.

teolemon:
We can do something special with the format you'd like @naivekook

VaiTon:
> @teolemon @VaiTon Hi!

Hey hi @naivekook and welcome!

> Is it still TBD? Can I take it?

Sure you can!

> also do we have another source of changelog?
> In current case we need to parse `CHANGELOG.md` via gradle task on each build to get the new changes and store them in application and it quite tricky because this file doesn't have stable formatting.

We could use the file used by the playstore maybe, I really do not know.

PR #3858, opened 2021-02-25
Fix Changelog.kt in release build

naivekook:
####  Description
I'm not able to build release build locally, so this is attempt to fix the issue. The main problem is the proguard obfuscates  these classes and jakson parser we cannot parse them in release build.

#### Related issues
#2870


naivekook:
@teolemon I'm not sure if it will work, but looks like problem is in proguard and kotlin nested data classes
cc @VaiTon

teolemon:
It's better to have very explicit names anyway 👍

PR #3880, opened 2021-03-09
ref: use hilt for dependency injection

VaiTon:


app/src/fdroid/java/openfoodfacts/github/scrachx/openfood/features/scan/ContinuousScanActivity.kt
```diff
@@ -346,12 +353,15 @@ class ContinuousScanActivity : AppCompatActivity() {
...
             }
-        }).also {
+        }, productRepository).also {
             it.loadAllergens { binding.callToActionImageProgress.visibility = View.GONE }
             it.loadAnalysisTags()
         }
     }
 
+    @Inject
+    lateinit var productRepository: ProductRepository
```
naivekook, line 363: lets move it at top of class?

naivekook:
wow nice work! I also have a local branch called `refactor-dagger-modules` 😄

Issue #3855, opened 2021-02-23
Re-apply the Matomo PR on top of the Kotlin code


teolemon:
Re-apply the Matomo PR on top of the Kotlin code.
We waited 6 months+ before merging and the Kotlin refactoring happened meanwhile.
As a result, rebase is complicated, so reapplying one by one the changes in the PR might be the right way to go.
- #3366

naivekook:
I can do it

VaiTon:
Sure you can! @naivekook

naivekook:
```
        //TODO: pass matomo url and id from properties, with different values depending on flavor
        tracker = TrackerBuilder.createDefault("https://example.com/piwik.php", 1)
```

@teolemon which urls should I use?

teolemon:
https://analytics.openfoodfacts.org/piwik.php

teolemon:
Site id =1

teolemon:
But we will need to create a separate one for Android

naivekook:
do we need also create another one for debug mode?

teolemon:
fixed

PR #3366, opened 2020-07-25
chore(stats): Matomo integration

philippeauriach:
**Type of Changes:** New feature

⚠️ Nothing is operational yet, as we don't have the Matomo server for now.

### TODO List:
- [x] ask the user at first open if he agreed to be tracked (default to false) (TODO)
- [x] in the preferences screen, allow to enable / disable tracking
- [x] in the preferences screen, allow disabling crash reporting

### Proposed changes
Integrates anonymous Matomo analytics tracking

**Views tracked:**
- scanner view
- products/edit/overview
- products/edit/ingredients
- products/edit/nutrition_facts

**Events tracked:**
- scanner scanned (with barcode as event name)
- scanner result card swiped up (with barcode as event name)
- allergen alert added (with allergen tag as event name)
- ingredient-analysis
  - enabled (with name of ingredient analysis)
  - disabled (with name of ingredient analysis)
- user-account
  - login
  - logout
  - login prompt (name: robotoff)
  - logged-in-after-prompt (name: robotoff)
- products :
  - created (with barcode as event name)
  - edited (with barcode as event name)
  - edited-ingredients-picture (added a product’s ingredients picture) (with barcode as event name)
  - compare-multiple : multiple products are being compared (triggered if products count > 1 with products count in event value)
  - compare-add : a product has been added for comparison
- shopping lists :
  - created
  - exported

### Screenshots
Screenshot_20200805-150506

Screenshot_1595668051

PR #3888, opened 2021-03-11
feat: matomo analytics

naivekook:
####  Description
I don't have permissions to update upstream branch, so created a new one.

#### Related issues
- #3855

#### Related PRs
- https://github.com/openfoodfacts/openfoodfacts-androidapp/pull/3366


app/src/main/java/openfoodfacts/github/scrachx/openfood/analytics/AnalyticsEvent.kt
```diff
@@ -0,0 +1,111 @@
+package openfoodfacts.github.scrachx.openfood.analytics
+
+sealed class AnalyticsEvent {
+
+    abstract val category: String
+    abstract val action: String
+    abstract val name: String?
+    abstract val value: Float?
+
+    data class ScannedBarcode(override val name: String) : AnalyticsEvent() {
```
VaiTon: Instead of overriding each time, let's create a constructon with the values and then initialize every class by calling super.
Like 
```kotlin
data class ScannedBarcode(name: String) : AnalysicsEvent(name, "scanner", "scanned", null);
```

app/src/main/java/openfoodfacts/github/scrachx/openfood/analytics/AnalyticsView.kt
```diff
@@ -0,0 +1,8 @@
+package openfoodfacts.github.scrachx.openfood.analytics
+
+enum class AnalyticsView(val path: String) {
+    Scanner("scanner"),
```
VaiTon, line 4: Let's use the common convention of having enums values as UPPERCASE:
https://stackoverflow.com/questions/3069743/coding-conventions-naming-enums

VaiTon (changes requested):
I will finish the review later this day

app/src/main/java/openfoodfacts/github/scrachx/openfood/analytics/AnalyticsView.kt
naivekook: np I'll change it, but I think in kotlin we should revisit some coding styles :)
good article https://arturdryomov.dev/posts/kotlin-enum-recipes/

VaiTon: I usually follow Android Studio hints, but yeah we should adopt a style.

VaiTon (approved):
Good job!

philippeauriach (approved):

naivekook:
@VaiTon can we ignore codefactor in this PR?

VaiTon:
@naivekook rebase and then good to merge!

naivekook:
@VaiTon done

VaiTon:
@naivekook sorry but still conflicts :(

Issue #3400, opened 2020-08-08
Add the option to use MLKit for scanning


teolemon:
Rationale: the current scanner is one of the top complaints ("can't scan", "too slow"). We're using a wrapper on top of zxing. zxing is maintenance only, while new scanners using on-device machine learning have appeared, and work far better, for far more people.
We could not use MLKit until now because it was bundled with Firebase which is a tracker. It's now fully unbundled, and while not opensource, it's not privacy invaside.
We should offer users the choice of getting that better scanner while keeping the fully free alternative.

How to: https://developers.google.com/ml-kit/vision/barcode-scanning/android

1. Preference to the embedded version that does not require Play Services
``` 
    dependencies {
      // ...
      // Use this dependency to bundle the model with your app
      implementation 'com.google.mlkit:barcode-scanning:16.0.2'
    }
```
2. Add a consent screen

New: Enhanced "MLKit" scanner 
We included a new option to more reliably scan barcodes. While this scanner works on your device, using machine learning, please be aware that it is a proprietary component provided by Google, governed by this privacy policy, and that some limited telemetry might be sent back to Google’s servers to improve their software. As noted in the MLKit terms, this won’t include any information about the products you scan and the telemetry is anonymized.
Choosing MLKit scanner implies that you accept those terms. 
Note: you can switch back at any time between the 2 scanners in the Settings.

MLKit Barcode Scanner |   | Classic Barcode Scanner

teolemon:
We should also see how we remove the dependancy for F-Droid, which builds the apps from the repository source.

kartikaysharma01:
Can you please assign this to me? @teolemon

PR #3900, opened 2021-03-19
Refactor MLKit camera functionality

naivekook:
####  Description
Having two `ContinuousScanActivity` causes a problems, because we have to support two version of the same code, this is too hard and we will always have bugs here. Even now we have small diff there because of hilt migration and we forgot to add matomo in second activity.

#### Related issues
#3855 
#3400 

#### Related PRs
none

#### Screenshots
none


VaiTon (approved):

VaiTon:
@naivekook rebase please

PR #3912, opened 2021-03-29
Automated DokkaDoc deployment to GitHub Pages using GitHub actions

teolemon:
Automated JavaDoc deployment to GitHub Pages using GitHub actions
https://github.com/marketplace/actions/github-pages-deploy-action

Output: https://openfoodfacts.github.io/openfoodfacts-androidapp/


avinal:
Hello @teolemon, I see you are trying to add Javadoc deployment to this project, seems like it is not working. Actually, I have not been updating my project lately, I can suggest some better alternatives and how to use them. Feel free to ask if you need help.

teolemon:
Hey @avinal, thank you so very much for reaching out proactively. 
Pointers on alternatives would be great.

avinal:
You are welcome @teolemon 
There are many options:
- If you prefer a more controlled and customized solution, then prefer writing your own GitHub Action, it isn't that hard, but it does come with a problem of an additional task to be taken care of. 
- The other option is to write a shell script and use an established deploy action such as https://github.com/peaceiris/actions-gh-pages, it provides a lot of control options and you don't have to maintain the action, just the script. I have upgraded CI of many open source orgs and used this, works nice.

I can open a PR but you might have to wait a day or two, I am a bit busy 😅

teolemon:
That would be awesome if you have time at some point 👍

avinal:
Hello @teolemon, I tried generating docs on my local machine, so far gradle config, dokka, and javadoc nothing seems to work. Maybe you have mixed other comment type with standard javadoc/kdoc comment style. If you can generate docs on your local machine please let me know the  commands. I will then simply write a wrapper Deploy Action.

teolemon:
It's working. Output: https://openfoodfacts.github.io/openfoodfacts-androidapp/

avinal:
Hello @teolemon, this is really nice.

.github/workflows/deploy-java-docs.yml
```diff
@@ -0,0 +1,36 @@
...
+            - name: Set up JDK 12
+              uses: actions/setup-java@v1
+              with:
+                java-version: 12
+            - name: Grant execute permission for gradlew
+              run: chmod +x gradlew
+            #- name: Build and run tests
+            #  run: ./gradlew build
+            - name: Generate API documentation
+              if: ${{github.event_name == 'push'}}
```
avinal: I think this line(27) and line number 30 both are unnecessary. You can remove lines 8 till 10 including 27 and 30 and it should work just the same.

.github/workflows/deploy-java-docs.yml
```diff
@@ -0,0 +1,36 @@
+    name: GitHub Pages Deploy Action
+    on:
+        # Branch name(s) which have the source file for site/documentation
+        push:
+            branches:
+                - "*"
```
avinal: In my opinion, you should avoid using wild cards and only update API documentation whenever there is a push in default branches (main or develop)

avinal (changes requested):

.github/workflows/deploy-java-docs.yml
teolemon: fixed

.github/workflows/deploy-java-docs.yml
teolemon: fixed, thank you 👍

PR #3926, opened 2021-04-05
Refactor network connectivity

naivekook:
####  Description
Remove deprecated access to network connectivity

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/allergensalert/AllergensAlertFragment.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/allergensalert/AllergensAlertFragment.kt
@@ -15,9 +15,7 @@
  */
 package openfoodfacts.github.scrachx.openfood.features.allergensalert
 
-import android.content.Context
 import android.content.SharedPreferences
-import android.net.ConnectivityManager
 import android.os.Bundle
 import android.util.Log
 import android.view.*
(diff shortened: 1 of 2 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/allergensalert/AllergensAlertFragment.kt
```diff
@@ -131,46 +130,51 @@ class AllergensAlertFragment : NavigationBaseFragment() {
...
-                        MaterialDialog.Builder(requireContext()).apply {
-                            title(R.string.title_dialog_alert)
-                            items(allergens.map { it.name })
-                            itemsCallback { _, _, position, _ ->
-                                productRepository.setAllergenEnabled(allergens[position].allergenTag, true)
-                                mAllergensEnabled!!.add(allergens[position])
-                                adapter.notifyItemInserted(mAllergensEnabled!!.size - 1)
-                                binding.allergensRecycle.scrollToPosition(adapter.itemCount - 1)
-                                matomoAnalytics.trackEvent(AnalyticsEvent.AllergenAlertCreated(allergens[position].allergenTag))
+                    .subscribe(
```
VaiTon: I prefer to use `doOnError` so that we can then use the single lambda on subscribe. Is there any reason to put it like this?

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/allergensalert/AllergensAlertFragment.kt
```diff
@@ -131,46 +130,51 @@ class AllergensAlertFragment : NavigationBaseFragment() {
...
-                        }
-                        .subscribe { _ ->
-                            mSettings.edit { putBoolean("errorAllergens", false) }
-                            adapter.allergens = mAllergensEnabled!!
-                            adapter.notifyDataSetChanged()
-                            updateAllergenDao()
-                            addAllergen()
-                            lt.success()
-                        }.addTo(disp)
+                        .subscribe(
```
VaiTon: Same as before

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/allergensalert/AllergensAlertFragment.kt
naivekook: actually no reason, will revert to previous approach

VaiTon (approved):

PR #3929, opened 2021-04-06
Move shared preferences to hilt

naivekook:
####  Description
In order to not depend on context everywhere and not to obtain SharedPreferences each time we trying to get language I moved it to hilt.


app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/LocaleHelper.kt
```diff
@@ -103,12 +104,12 @@ object LocaleHelper {
         return newContext
     }
 
-    fun setLanguageInPrefs(context: Context, language: String?): Context {
-        PreferenceManager.getDefaultSharedPreferences(context).edit {
+    fun setLanguageInPrefs(context: Context, language: String?, sharedPreferences: SharedPreferences): Context {
+        sharedPreferences.edit {
             putString(SELECTED_LANGUAGE, language)
             apply()
```
VaiTon:
```suggestion
```

`apply()` is useless when using `edit {}`

VaiTon (commented):
Why do you want to move shared preferences to hilt when we have to pass context too?

naivekook:
> Why do you want to move shared preferences to hilt when we have to pass context too?

this is the problem of current locale helper, my next PR will be refactor of LocaleHelper
I want move LocaleHelper in Hilt graph too and in this case you will not have to pass sharedprefs to it, because we will retrieve it from hilt in constructor

and because it will be in Hilt you will be able to use it in presenters and viemodels without having `context`

in code it will be like

```
@Inject
lateinit var localeHelper: LocaleHelper

val language = localeHelper.getLanguage()
```

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/LocaleHelper.kt
naivekook: right!

naivekook:
we can park this PR until I finish LocaleHelper refactor to see how it will look like

VaiTon:
@naivekook please rebase so we can see if it pass tests

VaiTon:
@naivekook rebase again please

teolemon:
rebase needed @naivekook

naivekook:
@VaiTon @teolemon rebased

VaiTon (commented):
Why did you remove the rules from androidTesting?

naivekook:
@VaiTon I didn't remove them , just moved to parent abstract class to make the hilt working automatically in child classes

PR #3935, opened 2021-04-08
Update codestyle to be aligned with official kotlin style

naivekook:
####  Description
After opening the project in AS Arctic Fox I'm getting the next warning about style. So I updated codestyle to be aligned with official kotlin style.

#### Screenshots
Screen Shot 2021-04-08 at 10 45 38

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -28,31 +28,8 @@
       <option name="JD_INDENT_ON_CONTINUATION" value="true" />
     </JavaCodeStyleSettings>
     <JetCodeStyleSettings>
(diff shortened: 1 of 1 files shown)


VaiTon:
I would prefer to postpone this until Artic Fox gets at least into beta.

naivekook:
Makes sense
Also if we want to automate code style or add some static analyze we can look at ktlint and detekt

VaiTon:
> Makes sense
> Also if we want to automate code style or add some static analyze we can look at ktlint and detekt

The Ideal solution would be to do this. Also using github actions to run this on PRs.

teolemon:
@naivekook @VaiTon Isn't that the case now ? (Artic Fox)

PR #3945, opened 2021-04-13
Refactor LocaleHelper

naivekook:
####  Description
This is second part of my refactoring. In this part I refactored `LocalHelper` and moved it to hilt.
Now we have `LocaleManager` which is responsible for applying locale to the app and `LocaleUtils` which is just util class to parse locale from language code and do some other helpful stuff.

#### Related PRs
#3929


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/HomeFragment.kt
```diff
@@ -197,25 +199,25 @@ class HomeFragment : NavigationBaseFragment() {
...
+                            taglineURL = tag.tagLine.url
+                            binding.tvDailyFoodFact.text = tag.tagLine.message
                             isLanguageFound = true
                         }
                     }
                     if (!isLanguageFound) {
-                        taglineURL = languages.last().tagLine.url
-                        binding.tvDailyFoodFact.text = languages.last().tagLine.message
+                        taglineURL = tagLines.last().tagLine.url
+                        binding.tvDailyFoodFact.text = tagLines.last().tagLine.message
```
teolemon: I've never understood the tvDailyFoodFact var name. Can we name it tagline as well, or something else ?

naivekook: sure

teolemon: @naivekook it's a rant from me, it's perfectly fine otherwise ;-)

naivekook:
will be super cool if someone can test it too

PR #3948, opened 2021-04-19
Handle lint warning for LocaleManager.kt

naivekook:
####  Description
Remove lint warning for deprecated method because we already handle this

#### Related PRs
#3945

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/LocaleManager.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/LocaleManager.kt
@@ -66,19 +66,20 @@ class LocaleManager @Inject constructor(
         } ?: currentLocale
     }
 
+    @Suppress("DEPRECATION")
     private fun changeAppLanguage(context: Context, locale: Locale): Context {
         var newContext = context
         Locale.setDefault(locale)
         currentLocale = locale
         val resources = newContext.resources
         val configuration = resources.configuration
-        if (Build.VERSION.SDK_INT >= 17) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/LocaleManager.kt
```diff
@@ -66,19 +66,20 @@ class LocaleManager @Inject constructor(
...
         currentLocale = locale
         val resources = newContext.resources
         val configuration = resources.configuration
-        if (Build.VERSION.SDK_INT >= 17) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
             configuration.setLocale(currentLocale)
             newContext = newContext.createConfigurationContext(configuration)
         } else {
             configuration.locale = currentLocale
+            resources.updateConfiguration(configuration, resources.displayMetrics)
```
VaiTon, line 81: I think that this should be placed outside the if. It should be executed in every case.

VaiTon (changes requested):

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/LocaleManager.kt
naivekook: according to the doc - no, I can do more testing on different emulators if you want

```
/**
     * Store the newly updated configuration.
     *
     * @deprecated See {@link android.content.Context#createConfigurationContext(Configuration)}.
     */
    @Deprecated
    public void updateConfiguration(Configuration config, DisplayMetrics metrics) {...}
```

VaiTon (approved):

PR #3954, opened 2021-04-23
Remove Sentry gradle plugin

### LINKED ISSUE #3830: Prevent Sentry from blocking the F-Droid build
Prevent Sentry from blocking the F-Droid build. Either by disabling it for F-Droid or making it work. We are 1 year (one year) late with the F-Droid version
https://gitlab.com/pierre.slamich/fdroiddata/-/jobs/1028085676#L2588

naivekook:
####  Description
Remove Sentry gradle plugin because we upload mapping with CLI.
https://docs.sentry.io/platforms/android/proguard/

#### Related issues
#3830

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -69,20 +69,19 @@ lane :release do
         skip_upload_screenshots: true
     )
 
+    FLAVOR = SharedValues::GRADLE_FLAVOR + SharedValues::GRADLE_BUILD_TYPE
+    sentry_upload_proguard(
+      org_slug: 'openfoodfacts',
+      project_slug: 'openfoodfacts-android',
+      android_manifest_path: './app/build/intermediates/merged_manifests/' + FLAVOR + '/out/AndroidManifest.xml'
+      mapping_path: './app/build/outputs/mapping/' + FLAVOR + '/mapping.txt',
+    )
+
     sentry_create_release(
         org_slug: 'openfoodfacts',
         project_slug: 'openfoodfacts-android',
         version: PACKAGE_NAME + '@' + versionName + '+' + versionCode,
         finalize: false # Release will be finalized by the "finalize_sentry" lane, when the version is in production
-    )
-
-    FLAVOR = SharedValues::GRADLE_FLAVOR + SharedValues::GRADLE_BUILD_TYPE
-    sentry_upload_proguard(
-      auth_token: ENV['SENTRY_AUTH_TOKEN'],
-      org_slug: 'openfoodfacts',
-      project_slug: 'openfoodfacts-android',
-      android_manifest_path: './app/build/intermediates/merged_manifests/' + FLAVOR + '/out/AndroidManifest.xml'
-      mapping_path: './app/build/outputs/mapping/' + FLAVOR + '/mapping.txt',
     )
 
     sentry_set_commits(
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -69,12 +69,20 @@ lane :release do
         skip_upload_screenshots: true
     )
 
-
     sentry_create_release(
         org_slug: 'openfoodfacts',
         project_slug: 'openfoodfacts-android',
         version: PACKAGE_NAME + '@' + versionName + '+' + versionCode,
         finalize: false # Release will be finalized by the "finalize_sentry" lane, when the version is in production
+    )
+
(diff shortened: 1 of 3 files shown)


naivekook:
ptal @teolemon @VaiTon

teolemon:
Probably @philippeauriach as well
Should we disable proguard ?
What does the Sentry plugin do ? Tag release ? Upload mappings ?
Disabling the Sentry plugin might not be a good idea beyond the F-Droid subcase, as it will require manual steps ?

naivekook:
Sentry gradle plugin uploads proguard mapping to sentry, but locally it didn't set up fully (missing `auth.token` for sentry)
I assume that on CI we upload the proguard mapping with CLI, so no need gradle plugin here.

VaiTon (commented):
Looks good to me. I would like to hear @philippeauriach

philippeauriach (changes requested):
Hello ! 

I don't think we send the mappings with the cli. When we release a build using Fastlane, we tell sentry there is a new release with the version that was just generated, and then send the corresponding commits, so that Sentry has the info.

In the Sentry dashboard, we do have symbolication, so the mappings are correctly sent. Indeed `auth.token` is not set, but there is a secret variable `secrets.SENTRY_AUTH_TOKEN` that was set by @teolemon, which is taken automatically by Sentry (injected in the gitub action).

I think that removing the plugin will cause the mappings to not be uploaded anymore.

Having said all that, the idea of sending the mappings using the fastlane plugin is good, as it will allow to remove the gradle plugin (as done in this PR). The only thing missing to accept this PR is therefore to implement the `sentry_upload_proguard` action in the release lane, something like here: https://github.com/openfoodfacts/openfoodfacts-androidapp/blob/60dd8d6edc975632023ad6a54248b946780418dc/fastlane/Fastfile#L73

See the following doc for the plugin usage: https://github.com/getsentry/sentry-fastlane-plugin#uploading-proguard-mapping-file

fastlane/Fastfile
```diff
@@ -69,14 +69,22 @@ lane :release do
...
     sentry_create_release(
         org_slug: 'openfoodfacts',
         project_slug: 'openfoodfacts-android',
         version: PACKAGE_NAME + '@' + versionName + '+' + versionCode,
         finalize: false # Release will be finalized by the "finalize_sentry" lane, when the version is in production
     )
 
+    FLAVOR = SharedValues::GRADLE_FLAVOR + SharedValues::GRADLE_BUILD_TYPE
+    sentry_upload_proguard(
+      auth_token: ENV['SENTRY_AUTH_TOKEN'],
```
philippeauriach: auth_token is not needed as passed in env variable (as in the other sentry calls)

philippeauriach (changes requested):
Thanks ! Just two small changes and we will be good to go !

fastlane/Fastfile
naivekook: @teolemon need your help, don't know how to inject token from CI

fastlane/Fastfile
```diff
@@ -69,14 +69,22 @@ lane :release do
...
+      org_slug: 'openfoodfacts',
+      project_slug: 'openfoodfacts-android',
+      android_manifest_path: './app/build/intermediates/merged_manifests/' + FLAVOR + '/out/AndroidManifest.xml'
+      mapping_path: './app/build/outputs/mapping/' + FLAVOR + '/mapping.txt',
+    )
+
```
philippeauriach: could this be set before `sentry_create_release`? Doing so we keep the same order of execution as before, when the upload was part of the build. It also means no sentry release will be created if we cannot upload the mappings, which is a good idea.

philippeauriach (changes requested):

fastlane/Fastfile
philippeauriach: I think there is no need to pass the variable, it is set globally with the correct name, therefore will be picked up by sentry. You can juste remove the parameter (same as other call to the sentry plugin, they are done without the parameter)

fastlane/Fastfile
naivekook: ah I thought that `sentry_create_release` is build the apk, now I see that it was done before by pure gradle task
so yes, will move it before `sentry_create_release`

fastlane/Fastfile
naivekook: noice! removed

philippeauriach (approved):

Issue #3949, opened 2021-04-21
Fix packager-codes URL


gargomoma:
**Describe the bug**
Hey, there is a problem when opening the 'Traceability Code' link while checking a product.

**To Reproduce**
1. Go to Product Page
2. Click on the 'Traceability Code' link
3. Browser will open
4. See error.

**Expected behavior**
Instead of:
https://world.openfoodfacts.org/packager-code/fsc-c014047

The generated URL is:
https://world.openfoodfacts.org/packager-code/https://world.openfoodfacts.org/packager-code/fsc-c014047

**Smartphone (please complete the following information):**
 - Device: Xiaomi Mix 2s (Polaris)
 - OS: Android 10
 - Version 3.6.3

teolemon:
https://github.com/openfoodfacts/openfoodfacts-androidapp/blob/1b3956398d8f81838d7a9f2fd67a0a2b53f5c906/app/src/main/java/openfoodfacts/github/scrachx/openfood/network/services/ProductsAPI.kt#L233

VaiTon:
>  `@GET("packager-code/{packager_code}.json")`

This is actually unused. I don't think that's the problem.

PR #3956, opened 2021-04-28
fix: wrong packager-codes url in SummaryProductFragment

VaiTon:
Fixes https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/3949


teolemon (approved):
the tag is the preferred way to build an URL. I'm not sure why we were querying the DB.

PR #4058, opened 2021-06-13
Non-working fix for the crash happening at startup #4055

teolemon:
https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/4055
Edit: the fix is not working


VaiTon:
@teolemon is this still necessary?

teolemon:
I've merged develop into release, meanwhile, I close this PR since it does nothing anyway

PR #4130, opened 2021-08-06
[4124] Fix uninitialized property exception in AllergensAlertFrament

naivekook:
####  Description
Fix the uninitialized property exception in AllergensAlertFrament.
This is more like hotfix - just check the property before calling its method. The idea case to understand why `destroy` happens before recycler initialization and refactor the code, but it requires time.

#### Related issues
#4124

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/allergensalert/AllergensAlertFragment.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/allergensalert/AllergensAlertFragment.kt
@@ -66,14 +66,13 @@ class AllergensAlertFragment : NavigationBaseFragment() {
     private lateinit var adapter: AllergensAdapter
     private val mSettings by lazy { requireActivity().getSharedPreferences("prefs", 0) }
     private val dataObserver by lazy { AllergensObserver() }
+    private val appLang: String by lazy { localeManager.getLanguage() }
 
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
         setHasOptionsMenu(true)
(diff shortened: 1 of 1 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/allergensalert/AllergensAlertFragment.kt
```diff
@@ -219,7 +220,7 @@ class AllergensAlertFragment : NavigationBaseFragment() {
     }
 
     companion object {
-        private val LOG_TAG = AllergensAlertFragment::class.simpleName
+        private const val LOG_TAG = "AllergensAlertFragment"
```
VaiTon: Why this?

VaiTon (changes requested):

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/allergensalert/AllergensAlertFragment.kt
naivekook: small optimization to not calculate the value in runtime and do it during compile time

VaiTon: All other classes do this. Also, I don't think it gives us any benefit since it's a one-time operation and in case of a refactor it should be updated. If you aren't strongly suggesting this I would revert.

naivekook: reverted, I removed LOG_TAG at all because it was unused

VaiTon (approved):

Issue #4139, opened 2021-08-07
Release fails because: allergensWrapper_SubElementsAreCorrectlyTagged FAILED


teolemon:
openfoodfacts.github.scrachx.openfood.models.entities.allergen.AllergenTest > allergensWrapper_SubElementsAreCorrectlyTagged FAILED
1294
    com.google.common.truth.ComparisonFailureWithFacts at AllergenTest.kt:69
1295

PR #4146, opened 2021-08-09
Fix AllergenTest

naivekook:
Fix allergen test

#### Related issues
#4139


teolemon (commented):
What did we change for the German allergen to break ?

naivekook:
It is a good question, from a test perspective looks like sort function in object parsing was changed

VaiTon (approved):

Issue #4117, opened 2021-07-30
Crash on 'Add products' in compare activity


teolemon:
It's not possible to use Compare Mode anymore. On an empty screen, you cannot click on "Add products" anymore.

naivekook:
fixed in https://github.com/openfoodfacts/openfoodfacts-androidapp/pull/4147

PR #4147, opened 2021-08-09
Fix Compare button click

naivekook:
Fix button click for Compare Mode 
#### Related issues
#4117


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/compare/ProductCompareActivity.kt
```diff
@@ -124,32 +124,29 @@ class ProductCompareActivity : BaseActivity() {
             photoReceiverHandler = PhotoReceiverHandler(sharedPreferences) {
                 productComparisonAdapter.onImageReturned(it)
             }
+        }
```
naivekook, line 127: just moved out click listener setup from `viewModel.products.observe` block

VaiTon (approved):
Thanks! 🎉

Issue #4140, opened 2021-08-07
The language names do not show anymore (edit mode, language picker)


teolemon:
The language names do not show anymore (edit mode, language picker)
![image](https://user-images.githubusercontent.com/1689815/128605448-85952250-9958-4660-a0d6-4202d0712847.png)

teolemon:
See #2958 with a mockup
![](https://user-images.githubusercontent.com/1689815/71359523-f6885380-258c-11ea-8047-630a80b514ba.png)

naivekook:
I can take this, which variant from mockups is preferable? Just name or name + code?

teolemon:
I like the implementation from the image manager with name plus code

teolemon:
Fixed. Thanks @naivekook

PR #4149, opened 2021-08-09
Show language name in picker for product edit screen

naivekook:
Show language name in picker for product edit screen

#### Related issues
#4140


VaiTon:
Kudos! 🎉

PR #4153, opened 2021-08-11
ref: Remove DataBinding from project

naivekook:
###  Description
I took the initiative of making this change without final resolution in the slack thread https://openfoodfacts.slack.com/archives/C02L15QLA/p1628503633025900.

As I said in the slack all our new code was written without data-binding approach, so looks like we depend on this library only for legacy code. By removing this library we can use more clean XML files and decrease build time from 2.29min to 1.26min(!!). I didn't make any measuring of apk file but think it is decreased too.

This PR removes data-binding from the project and introduces view-binding instead. 
There are a lot of changes inside XML files, but mostly this is formatting (because of removing layout and data tags) and adding `tools:ignore="ContentDescription"` to ImageView to prevent lint from warning.
Also, some fragments were fixed because they used to use data-binding before.

#### With data binding
https://scans.gradle.com/s/gv47vexa7beck/performance/build

#### Without data binding
https://scans.gradle.com/s/ggg6psw3cdnca/performance/build

#### With data binding
Screenshot 2021-08-11 at 08 20 26

#### Without data binding
Screenshot 2021-08-11 at 08 20 34

#### Link to the APK artefact
[APK for OFF](https://github.com/openfoodfacts/openfoodfacts-androidapp/suites/3493821138/artifacts/83204688)


VaiTon:
@naivekook I really don't know. 

Data binding is used in some code which is not legacy and I don't know if we should remove it completely. The reduction in the compilation time is really good but it's really useful to put some logic code in the xml file instead of cluttering up the Activity/Fragment.

Also, with the introduction of Composable views, which I would like to introduce to this project when we remove the API requirement for older version, will completely change how we work with views.

@teolemon any view?

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/analyticsusage/AnalyticsUsageDialogFragment.kt
```diff
@@ -45,10 +48,15 @@ class AnalyticsUsageDialogFragment : BottomSheetDialogFragment() {
...
+    }
+
     private fun saveAnalyticsReportingPref(value: Boolean) {
         sharedPreferences
-                .edit()
-                .putBoolean(getString(R.string.pref_analytics_reporting_key), value)
-                .apply()
+            .edit()
+            .putBoolean(getString(R.string.pref_analytics_reporting_key), value)
+            .apply()
```
VaiTon:
```suggestion
            .edit { putBoolean(getString(R.string.pref_analytics_reporting_key), value) }
```

And import the required extension.

VaiTon (changes requested):
Partial review, will complete later

VaiTon:
@naivekook also, it doesn't compile

naivekook:
@VaiTon  sure it is always a trade-off
My opinion here is not to mix approaches inside the code. For example, you have a bug, you go to Fragment and try to find and fix it, but looks like all works fine, but suddenly, you understand, that some variable is injected into XML directly by data-binding and it causes this bug.
And how many fragments do we have? about 20? and only 3-4 uses the data-binding, so no such big price for improving the build time and code sustainability
By the way, you don't use both java and kotlin for new code right? :)

> with the introduction of Composable views

wow that's awesome! love the Jetpack Compose
but with introducing the compose this in another +1 to get rid of data-binding ;)

> also, it doesn't compile

I tested only offFdroidDebug variant, will fix all other variants and tests if we will decide to ship this PR


anyway I'm not pushing much, just want to bring another point of view and possible improvement :)

VaiTon:
@naivekook ok, you convinced me. Let's do this :)

teolemon:
I was about to say that I had no opinion on the matter ;-)

naivekook:
@VaiTon review it again, please
I tested FDroid and Playstore versions on the device and looks like it is all fine

VaiTon (approved):

teolemon:
@VaiTon @naivekook safe to merge ?

naivekook:
@teolemon merge it after release please

teolemon:
ok, noted.

naivekook:
I'm working on MacBook air right now and the build time is awful, can we consider merging it right now?

VaiTon:
@teolemon ?

Issue #4083, opened 2021-07-01
It seems impossible to compare more than 2 products anymore


teolemon:
It seems impossible to compare more than 2 products anymore. It seemed to crash at scan time and go back to comparison mode. 
I looked at Sentry, and I couldn't find any related trace. Will update if it gets uploaded eventually. 
The issue was happening for at least 3 different products.

teolemon:
Still happening on latest dev (11 july 2021)

naivekook:
should be fixed in #4191
will check it again and write test on it

VaiTon:
Fixed by #4191

teolemon:
I'm still unable to compare more than 2 products ?

naivekook:
Hm very very strange, will look at it again tomorrow

naivekook:
@teolemon checked last develop - all works
can you describe the steps to reproduce?

PR #4191, opened 2021-08-27
ref: product compare screen

naivekook:
####  Description
As I said in my last PR I'm going to refactor CompareProductActivity and here are the results!
The main idea of this PR is to move the logic to the viewmodel and cover it with tests. Now I hope will be harder to break something inside compare screen. :)

What was done:
- Simplified the ContinuousScanActivity by moving related logic to ProductCompareViewModel
- Moved comparison logic from fragment to view model
- Wrote tests for ProductCompareViewModel
- Tweak a bit dependencies
- Added CoroutineDispatchers and some utils classes for testing

#### Screenshots
video link
https://drive.google.com/file/d/1SW9ILafVQ51YOtv6LpO-hs1cWC55f5UR/view?usp=sharing

#### Test APK
https://github.com/openfoodfacts/openfoodfacts-androidapp/suites/3612089427/artifacts/87254330

#### Part of
- #4209


app/build.gradle.kts
```diff
@@ -68,7 +67,6 @@ dependencies {
     implementation("androidx.multidex:multidex:2.0.1")
     implementation("androidx.viewpager2:viewpager2:1.0.0")
     implementation("androidx.legacy:legacy-support-v4:1.0.0")
-    implementation("androidx.lifecycle:lifecycle-extensions:2.2.0")
```
naivekook, line 71: this is deprecated library, don't needed anymore

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/compare/ProductCompareViewModel.kt
```diff
@@ -1,38 +1,65 @@
...
+import openfoodfacts.github.scrachx.openfood.utils.SingleLiveEvent
 import javax.inject.Inject
 
 @HiltViewModel
 class ProductCompareViewModel @Inject constructor(
     private val productRepository: ProductRepository,
-    private val localeManager: LocaleManager
+    private val localeManager: LocaleManager,
+    private val matomoAnalytics: MatomoAnalytics,
+    private val coroutineDispatchers: CoroutineDispatchers
```
naivekook, line 27: we have to use CoroutineDispatchers if we want coroutines work in tests

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/compare/ScanProductActivityContract.kt
```diff
@@ -0,0 +1,23 @@
+package openfoodfacts.github.scrachx.openfood.features.compare
+
+import android.app.Activity
+import android.content.Context
+import android.content.Intent
+import androidx.activity.result.contract.ActivityResultContract
+import openfoodfacts.github.scrachx.openfood.features.scan.ContinuousScanActivity
+import openfoodfacts.github.scrachx.openfood.models.Product
+
+class ScanProductActivityContract : ActivityResultContract<Unit, Product?>() {
```
naivekook, line 12: handy contract to work with ContinuousScanActivity instead of deprecated `startActivityForResult`

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/scan/ContinuousScanActivity.kt
```diff
@@ -245,19 +246,8 @@ class ContinuousScanActivity : BaseActivity(), IProductView {
...
-                        val productsToCompare = intent.extras!!.getSerializable(ProductCompareActivity.KEY_PRODUCTS_TO_COMPARE) as ArrayList<Product>
-                        if (product in productsToCompare) {
-                            putExtra(ProductCompareActivity.KEY_PRODUCT_ALREADY_EXISTS, true)
-                        } else {
-                            productsToCompare += product
-                            matomoAnalytics.trackEvent(AnalyticsEvent.AddProductToComparison(product.code))
-                        }
-                        putExtra(ProductCompareActivity.KEY_PRODUCTS_TO_COMPARE, productsToCompare)
-                        addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP)
-                    })
```
naivekook, line 260: all this logic moved to CompareProductsViewModel

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/SingleLiveEvent.kt
```diff
@@ -0,0 +1,77 @@
...
+ *
+ *
+ * This avoids a common problem with events: on configuration change (like rotation) an update
+ * can be emitted if the observer is active. This LiveData only calls the observable if there's an
+ * explicit call to setValue() or call().
+ *
+ *
+ * Note that only one observer is going to be notified of changes.
+ */
+class SingleLiveEvent<T> : MutableLiveData<T>() {
```
naivekook: this is live data that emits only once

app/src/test/java/openfoodfacts/github/scrachx/openfood/utils/LiveDataTestUtil.kt
```diff
@@ -0,0 +1,45 @@
...
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+
+/**
+ * Gets the value of a [LiveData] or waits for it to have one, with a timeout.
+ *
+ * Use this extension from host-side (JVM) tests. It's recommended to use it alongside
+ * `InstantTaskExecutorRule` or a similar mechanism to execute tasks synchronously.
+ */
+fun <T> LiveData<T>.getOrAwaitValue(
```
naivekook: test helper to test livedata

naivekook:
looks like we need to turn off SonarCloud for test classes

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/SingleLiveEvent.kt
VaiTon: We could use EventBus for events

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/compare/ProductCompareActivity.kt
```diff
@@ -160,27 +81,79 @@ class ProductCompareActivity : BaseActivity() {
...
-        const val KEY_PRODUCT_FOUND = "product_found"
         const val KEY_COMPARE_PRODUCT = "compare_product"
-        const val KEY_PRODUCT_ALREADY_EXISTS = "product_already_exists"
 
         @JvmStatic
         fun start(context: Context, product: Product) {
             context.startActivity(Intent(context, ProductCompareActivity::class.java).apply {
-                putExtra(KEY_PRODUCT_FOUND, true)
-                putExtra(KEY_PRODUCTS_TO_COMPARE, arrayListOf(product))
+                putExtra(KEY_PRODUCTS_TO_COMPARE, product)
```
VaiTon, line 171: We use this here but we do not retrieve the value of this inside the activity. What if i launch "compare" from a product screen?

VaiTon (changes requested):

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/SingleLiveEvent.kt
naivekook: right, but I don't want to use it here because
- this VM already uses LiveData, SingleLiveEvent is a common approach for such cases with LiveData inside (actually I copied it from google with copyrights), so it will be right to say that we solve the problem without introducing new dependencies
- I don't want to use EventBus here because right now we have direct relation between activity and view model, and activity just reads the value from it. With introducing EventBus it will be harder relation like view model sends some data to EventBus, activity listen to the event of some type from EventBus, so in the result, we introduced new logical instance here and it becomes more complicated
- EventBus is quite hard to support imo

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/compare/ProductCompareActivity.kt
naivekook: good callout, will check it again

naivekook: fixed

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/CoroutineDispatchers.kt
```diff
@@ -0,0 +1,21 @@
...
+import javax.inject.Inject
+
+interface CoroutineDispatchers {
+
+    fun main(): CoroutineDispatcher
+    fun io(): CoroutineDispatcher
+    fun default(): CoroutineDispatcher
+}
+
+class CoroutineDispatchersImpl @Inject constructor() : CoroutineDispatchers {
```
VaiTon, line 14: Why are we using this? Why don't we use default dispatchers?

VaiTon (commented):
I don't like the SingleLiveEvent approach a lot. What about using coroutines? Something like StateFlow?

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/SingleLiveEvent.kt
VaiTon: > EventBus is quite hard to support imo

Why?

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/compare/ProductCompareViewModel.kt
VaiTon: We could use runBlocking. We already use it in some tests. I tried using runTestBlocking in the past but it errored.

VaiTon:
@naivekook what about https://developer.android.com/topic/libraries/architecture/viewmodel#sharing using `MutableStateFlow`

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/compare/ProductCompareViewModel.kt
naivekook: I tried to use runBlocking in my tests but it didn't work well, that's why I inject dispatchers wrapper here

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/SingleLiveEvent.kt
naivekook: because over time it became like an octopus with many tentacles and exists everywhere, it is hard to understand which class uses a particular event from EventBus, and changing some part of logic in one place can lead us to errors in unexpected places
and also you have to always control for subscribing\unsubscribing from it

naivekook:
@VaiTon coroutines works for me! will replace the LiveData with stateflow

app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/CoroutineDispatchers.kt
naivekook: for mocking default dispatchers in test, runTestBlocking didn't work for me, maybe because we have rxjava2->coroutine mapper inside

teolemon:
@naivekook @VaiTon I created a meta-bug for the Compare mode:
- #4209

VaiTon:
@naivekook LGTM! ✅

naivekook:
@VaiTon Cool! you can merge it, it also fixes the issue #4083

Issue #4224, opened 2021-09-05
Clicking on the ingredient analysis crashes the view


teolemon:
Clicking on the ingredient analysis crashes the view

**To Reproduce**
1. Go to a product with analysis complete
2. Click on one of the analysis chips
3. It crashes

**Stacktrace**

- [Sentry issue](https://sentry.io/organizations/openfoodfacts/issues/2627439752/?referrer=github_integration)
```
Fragment IngredientsWithTagDialogFragment{6db927a} (959a4d2f-56bf-480a-b358-e77d2e6c7a27) has not been attached yet.
mechanism
UncaughtExceptionHandler
handled
false
```

```
androidx.fragment.app.Fragment in getChildFragmentManager at line 1075
… hub.scrachx.openfood.features.product.view.summary.SummaryProductFragment$showAnalysisTags$2$1$1 in invoke at line 622
… hub.scrachx.openfood.features.product.view.summary.SummaryProductFragment$showAnalysisTags$2$1$1 in invoke at line 616
… ood.features.product.view.summary.IngredientAnalysisTagsAdapter$IngredientAnalysisTagsViewHolder in onClick at line 76
android.view.View in performClick at line 7441
android.view.View in performClickInternal at line 7418
android.view.View in access$3700 at line 835
android.view.View$PerformClick in run at line 28674
android.os.Handler in handleCallback at line 938
android.os.Handler in dispatchMessage at line 99
android.os.Looper in loopOnce at line 201
android.os.Looper in loop at line 288
android.app.ActivityThread in main at line 7822
java.lang.reflect.Method in invoke
com.android.internal.os.RuntimeInit$MethodAndArgsCaller in run at line 548
com.android.internal.os.ZygoteInit in main at line 1003
```
**Screenshots**

https://user-images.githubusercontent.com/1689815/132131909-98e4b4b1-2c67-4b41-8be7-1538467ecbae.mp4

PR #4227, opened 2021-09-06
Fix clicking on the ingredient analysis view

naivekook:
####  Description
Fix clicking on the ingredient analysis view

#### Related issues
- #4224

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/summary/SummaryProductFragment.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/summary/SummaryProductFragment.kt
@@ -36,6 +36,7 @@ import androidx.core.net.toUri
 import androidx.core.text.bold
 import androidx.core.text.buildSpannedString
 import androidx.core.text.inSpans
+import androidx.core.view.isVisible
 import androidx.core.view.updatePadding
 import androidx.lifecycle.lifecycleScope
 import androidx.recyclerview.widget.LinearLayoutManager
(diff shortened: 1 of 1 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/summary/SummaryProductFragment.kt
```diff
@@ -603,24 +604,20 @@ class SummaryProductFragment : BaseFragment(), ISummaryProductPresenter.View {
...
                         setOnItemClickListener { view, _ ->
-                            IngredientsWithTagDialogFragment.newInstance(
+                            val fragment = IngredientsWithTagDialogFragment.newInstance(
                                 product,
                                 view.getTag(R.id.analysis_tag_config) as AnalysisTagConfig
-                            ).run {
-                                onDismissListener = { filterVisibleTags() }
-                                show(childFragmentManager, "fragment_ingredients_with_tag")
-                            }
+                            )
```
VaiTon, line 618: What changes? Could you explain with a comment so that we do not repeat the same mistake?

VaiTon (approved):

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/summary/SummaryProductFragment.kt
naivekook: sure!
the tricky part here is in how `run` works - it creates a lambda and passes the fragment as a parameter inside, other words it creates a scoped place with this fragment only
so the code looks similar w/o `run` but works differently, the fragment has a complex lifecycle and inside `run` lambda it doesn't pass through all method and doesn't attach to the child fragment manager (because lambda has own scope)
so when you call `show` fragment manager cant show it

in a result inside `run` we can setup a listener and it will work, but showing doesn't work

teolemon:
can we merge @naivekook @VaiTon ?

naivekook:
@teolemon yeap, ship it

Issue #4214, opened 2021-09-03
Simple scan activity


naivekook:
**Is your feature request related to a problem? Please describe.**
Right now we use ContinuousScanActivity everywhere we need to scan the product, it leads to the situation when we have different pieces of logic for different use cases in one activity
for example, in comparison mode when you open this activity and scan the product you can see for half a sec the bottom view with product details (timing 0:08 in this video https://drive.google.com/file/d/1SW9ILafVQ51YOtv6LpO-hs1cWC55f5UR/view) and under the hood it also does some logic which is unnecessary for the current use case
also in this scan activity you can open other different screens because the bottom navigation bar is available here, that is not very good UX imo, because the user quitting from flow of adding the product

**Describe the solution you'd like**
I want to create an additional SimpleScanActivity that will just do the scan and return the result back. Single responsibility and simple to use.

**Describe alternatives you've considered**
Refactor ContinuousScanActivity to support all scan modes, but it is quite hard and requires a lot of time + we have a risk to break the build for users.

naivekook:
@VaiTon @teolemon needs your opinion

VaiTon:
How would you avoid duplicating things?

naivekook:
@VaiTon do you mean any particular thing or duplication in general?

My thoughts that it will be duplication only in layout for barcode scanner and some logic which retrieve barcode from scanner and call API for product.

This is tradeoff and on one side we have let's say "feature duplication" but on other side we have better UX, less API calls, more simple and straightforward logic. Plus we can write tests on this.

VaiTon:
Ok then, you can proceed! @naivekook

PR #4230, opened 2021-09-08
Add repository to work with camera prefs

naivekook:
####  Description
Extract all logic related to changing camera prefs to a separate repository. This is needed to use the same settings for ContinuousScan and SimpleScan features.

#### Related issues
- #4214

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/scan/ContinuousScanActivity.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/scan/ContinuousScanActivity.kt
@@ -31,7 +31,6 @@ import android.widget.Toast
 import androidx.annotation.StringRes
 import androidx.appcompat.widget.PopupMenu
 import androidx.core.content.ContextCompat
-import androidx.core.content.edit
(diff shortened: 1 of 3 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/scan/ContinuousScanActivity.kt
```diff
@@ -223,7 +225,7 @@ class ContinuousScanActivity : BaseActivity(), IProductView {
                         .apply { setGravity(CENTER, 0, 0) }
                         .show()
 
-                    Log.w(LOG_TAG, err.message, err)
```
VaiTon, line 226: Why?

naivekook: because LOG_TAG uses explicit unwrap here, I'm trying to avoid using `!!` everywhere just because it requires always think "is it safe to use !! here or not? does it crash on production?"
and we use this tag only in one place so I changed it to a simple string

naivekook:
merged within #4236

PR #4294, opened 2021-10-05
For #4247: Perform layout updates on the main thread.

mcarare:
####  Description
Move performing the layout updates to the main thread. 
(Using await() in the try block used the coroutine scope in the catch too.)

#### Related issues
#4247

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/edit/ProductEditActivity.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/edit/ProductEditActivity.kt
@@ -527,7 +527,7 @@ class ProductEditActivity : BaseActivity() {
             try {
                 productsApi.performOCR(code, imageField).await()
             } catch (err: Exception) {
-                ingredientsFragment.hideOCRProgress()
+                withContext(Main) { ingredientsFragment.hideOCRProgress() }
                 if (err is IOException) {
                     val view = findViewById<View>(R.id.coordinator_layout)
                     Snackbar.make(view, R.string.no_internet_unable_to_extract_ingredients, LENGTH_INDEFINITE)


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/edit/ProductEditActivity.kt
```diff
@@ -527,7 +528,7 @@ class ProductEditActivity : BaseActivity() {
             try {
                 productsApi.performOCR(code, imageField).await()
             } catch (err: Exception) {
-                ingredientsFragment.hideOCRProgress()
+                MainScope().launch { ingredientsFragment.hideOCRProgress() }
```
naivekook: I think `withContext` operator is better here
`withContext(Dispatchers.MAIN) { ingredientsFragment.hideOCRProgress() }`

mcarare: Why? Do we need to wait for a result from `hideOCRProgress`?

naivekook: this is not about getting the result of the lambda
`MainScope` is not attached to the lifecycle of an activity, so **potentially** we can have a crash here

mcarare: Not really sure what crash you can get from this. And how is `withContext ` better for the reason you mentioned.

naivekook: `withContext` runs within the `lifecycleScope` (line 410) and the job will be canceled when the activity gets destroyed.
MainScope doesn't know about activity lifecycle and you have to cancel it when activity destroys. In this part of code you create a new job, which will run in scope, which is not lifecycle aware and can crash because `ingredientsFragment` not exists anymore.
I understand a very small chance that this will happen but this is not a good reason to use MainScope here.

mcarare: You are aware the changes in the suspend block are for a fragment in another `lifecycleScope` ?

naivekook: Yes, and this fragment is attached to this activity. What's the question?

mcarare: Nevermind. :)

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/edit/ProductEditActivity.kt
```diff
@@ -35,6 +35,7 @@ import com.google.android.material.snackbar.Snackbar
 import dagger.hilt.android.AndroidEntryPoint
 import kotlinx.coroutines.Dispatchers.IO
 import kotlinx.coroutines.Dispatchers.Main
+import kotlinx.coroutines.MainScope
```
naivekook: unused import?

naivekook (approved):

mcarare:
@naivekook does this need any other changes in order to land?

naivekook:
> @naivekook does this need any other changes in order to land?

Hey!
I help the team with code review, but the last word on @VaiTon and @teolemon.

teolemon:
I fully trust @naivekook @VaiTon on the technical part. @naivekook feel free to merge 👍

Issue #4267, opened 2021-09-28
Trying to add a non existing product to a comparison creates a confusing message


teolemon:
**Description**
Trying to add a non existing product to a comparison creates a confusing message

**To Reproduce**
1. Start a comparison
2. Scan a book
3. Boom

**Expected behavior**
"This product can't be added to the comparison because it does not exist in Open Food Facts. Would you like to add it ?"

**Screenshots**
![image](https://user-images.githubusercontent.com/1689815/135141681-1ee45eab-a198-4d3e-92b8-467825bc4e5a.png)

TomerPacific:
Hi @teolemon , I'd like to help out.

naivekook:
just need to replace `txtDialogsContentPowerMode` with `txtDialogsContent` on line 190 in ProductCompareActivity
https://github.com/openfoodfacts/openfoodfacts-androidapp/blob/dd598272e6c32da7a816dabcaad2480ea1887591/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/compare/ProductCompareActivity.kt#L190

PR #4299, opened 2021-10-08
Fix comparison dialog text

naivekook:
####  Description
Changed the text to `The product you scanned does not exist yet in Open Food Facts. Please add it.`

#### Related issues
- #4267


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/compare/ProductCompareActivity.kt
```diff
@@ -187,7 +187,7 @@ class ProductCompareActivity : BaseActivity() {
 
     private fun showProductNotFoundDialog() {
         MaterialAlertDialogBuilder(this)
-            .setMessage(R.string.txtDialogsContentPowerMode)
+            .setMessage(R.string.txtDialogsContent)
```
teolemon: For reference: 
```<string name="txtDialogsContent">The product you scanned does not exist yet in Open Food Facts. Please add it.</string>```
We could probably rename the string, it's not very explicit

naivekook: yeah agree, changed

teolemon (approved):

Issue #4120, opened 2021-08-01
Allergen alert set in the previous version of the app (production) disappear in the beta


teolemon:
An allergen alert set in the previous version of the app (production) disappears in the beta.

**Steps to reproduce:**
- I installed Open Food Facts as a normal user
- I added a gluten alert
- I tested that it worked
- I joined the beta (which is `develop`)
- I saw the gluten alert had disappeared
- I added a new one
- It worked again

**Potential causes**
```
2021-09-12 17:58:39.197 21409-21409/? D/Sentry: io.sentry.auto-init read: false
2021-09-12 17:58:39.197 21409-21409/? I/Sentry: Retrieving auto-init from AndroidManifest.xml
2021-09-12 17:58:39.211 21409-21409/? I/greenDAO: Migrating schema from version 18 to 22
2021-09-12 17:58:39.211 21409-21409/? E/MIGRATE VERSION: 19
2021-09-12 17:58:39.213 21409-21409/? E/MIGRATE VERSION: 20
2021-09-12 17:58:39.215 21409-21409/? E/MIGRATE VERSION: 21
2021-09-12 17:58:39.216 21409-21409/? E/MIGRATE VERSION: 22
2021-09-12 17:58:39.288 21409-21438/? I/Sentry: sentry-debug-meta.properties file was not found.
```
- [ ] [The file where migrations are done](https://github.com/openfoodfacts/openfoodfacts-androidapp/blob/develop/app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/OFFDatabaseHelper.kt)
- [ ] https://github.com/openfoodfacts/openfoodfacts-androidapp/commit/0e06205848ed028061e346c9c1d23acca9a3c6ad#diff-5659840340e38dfaad346e16d4456a24a326f02a343e51331c7cf2d618b4da9e
- [ ] https://github.com/openfoodfacts/openfoodfacts-androidapp/commit/c558f1b29de9a8fcfaebb26b4ff62d41de505abd#diff-5659840340e38dfaad346e16d4456a24a326f02a343e51331c7cf2d618b4da9e

### Date of the production release (last known working)
Last updated Jan 14 8:16 PM
So roughly release 3.6.3 @ https://github.com/openfoodfacts/openfoodfacts-androidapp/commit/b691a6e9cae94dadb8659f51fae01d63c0e0f153

teolemon:
[...]

PR #4325, opened 2021-10-24
Replacing allergens should keep their "enabled" statuses

g123k:
The issue #4120 come from the fact that when taxonomies are refreshed, all allergens are replaced.
BUT an allergen may have its "enabled" column at `true`, which means it's also an alert.
Here, I copy the "enabled" value to be sure to not lose the previous state


teolemon:
✅  LGTM, I'm pushing to internal, and testing if the fix works.

PR #4363, opened 2021-11-18
ref: use coroutines instead of rxjava in repositories

VaiTon:
ref: "services" renamed to "repositories" as per android docs recommendation


teolemon:
> Task :app:compileObfPlaystoreDebugKotlin FAILED
e: /home/runner/work/openfoodfacts-androidapp/openfoodfacts-androidapp/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/MainActivity.kt: (938, 98): Unresolved reference. None of the following candidates is applicable because of receiver type mismatch: 
public suspend fun CompletableSource.await(): Unit defined in kotlinx.coroutines.rx2

Issue #2867, opened 2019-08-20
Default units for vitamins, fats and micronutrients.


Slawek234:
When we add a Vitamin, fats, macro- microelements the default unit should be selected.

Current situation. The default unit is gram.
<a href='https://postimg.cc/K4cV6y9x' target='_blank'><img src='https://i.postimg.cc/K4cV6y9x/Screenshot-2019-08-20-13-28-35.png' border='0' alt='Screenshot-2019-08-20-13-28-35'/></a>

After entering the automatic units for the different nutrients.
<a href='https://postimg.cc/LJmZTN9C' target='_blank'><img src='https://i.postimg.cc/LJmZTN9C/Screenshot-2019-08-20-13-28-12.png' border='0' alt='Screenshot-2019-08-20-13-28-12'/></a>

hmuhammadazeem:
Is there any database that provides units according to the nutrients? I'm willing to work on this.

Slawek234:
On this page https://eur-lex.europa.eu/legal-content/EN/TXT/HTML/?uri=CELEX:32011R1169&from=en , in Annex 13 I found a list of added vitamins and minerals with units.

Vitamin A (μg)
Vitamin D (μg)
Vitamin E (mg)
Vitamin K (μg)
Vitamin C (mg)
Thiamin (mg)
Riboflavin (mg)
Niacin (mg)
Vitamin B6 (mg)
Folic acid (μg)
Vitamin B12 (μg)
Biotin (μg)
Pantothenic acid (mg)
Potassium (mg)
Chloride (mg)
Calcium (mg)
Phosphorus (mg)
Magnesium (mg)
Iron (mg)
Zinc (mg)
Copper (mg)
Manganese (mg)
Fluoride (mg)
Selenium(μg)
Chromium (μg)
Molybdenum (μg) 
Iodine (μg)
-------------------------
caffeine (mg)

teolemon:
@Slawek234 can you file on https://github.com/openfoodfacts/openfoodfacts-server ?
I'd rather provide the nutrient translations as a downloadable file we can update on the fly. Also, the European Commission is making its translation memories available as opendata, so we'll be able to build on that.

VaiTon:
I think this is already implemented in openfoodfacts-server.

teolemon:
https://static.openfoodfacts.org/data/taxonomies/nutrients.json
```
[...]

PR #4373, opened 2021-11-24
Fix issue #2867: Honor the default unit from the nutrient taxonomy

machiav3lli:
####  Description
A first Iteration of our suggested solution in the discussion under #2867 .
For this one we focused on the logic managing binding each nutrient with a default value and this saved statically in a map.
A smal test is also included that checks if the map still binds the true default values for some 4 cherry-picked nutrients.

#### Related issues
#2867. This solves the issue fully, but another variant is also possible. The two variants are:
- Use a static map to bind (like in this first iteration)
    - Have to be maintained (Should be in the future new nutrients to be considered).
    - More efficient considering offline usage of the app.
- Use OFF JSON data to get the default value of each nutrient
    - Gets maintained on the server side.
    - Less efficient considering calls etc.

We would be interested to hear the opinion of the community+maintainer(s)…


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/edit/nutrition/ProductEditNutritionFactsFragment.kt
```diff
@@ -712,17 +711,25 @@ class ProductEditNutritionFactsFragment : ProductEditFragment() {
         }
 
     private fun displayAddNutrientDialog() {
-        val nutrients = resources.getStringArray(R.array.nutrients_array)
+        val nutrientsDefUnits = resources.getStringArray(R.array.nutrients_array).zip(PARAMS_OTHER_NUTRIENTS_DEFAULT_UNITS.keys)
+        val filteredNutrients = resources.getStringArray(R.array.nutrients_array)
             .filterIndexed { index, _ -> index !in usedNutrientsIndexes }
-            .sortedWith(Collator.getInstance(Locale.getDefault()))
+            .sortedBy { it.lowercase() }
```
VaiTon: Why did you remove this? By using `sortedBy { it.lowercase() }` we're not considering all Unicode languages.

machiav3lli: Sorry, was a part of one of a previous implementation idea that we forgot to clean up. But now it's back

VaiTon:
@machiav3lli tests are failing

machiav3lli:
@VaiTon yeah, I've forgotten the import

VaiTon:
@machiav3lli could you fix it so that tests work?

machiav3lli:
> @machiav3lli could you fix it so that tests work?

I've already fixed before I wrote the note about the missing import and it builds just fine on my side. The UnitTest that is failing is:
_openfoodfacts.github.scrachx.openfood.network.ProductsAPITest > productByAdditive_e301_productsFound_
```java
@Test
    fun productByAdditive_e301_productsFound() {
        val fieldsToFetchFacets = "brands,product_name,image_small_url,quantity,nutrition_grades_tags"
        val response = prodClient.getProductsByAdditive("e301-sodium-ascorbate", fieldsToFetchFacets).blockingGet() as Search
        assertThat(response).hasFoundProducts()
    }
```

 I guess it could be an error/change on the server side.

Lint OFF seems also to be failing, but all the mentioned errors doesn't relate to the classes touched by this PR in anyway and for the most part are (string) formatting issues.

Please check those and tell me if we should try to fix those or if it's maybe someone else's responsibility (server for UnitTest, OFF-maintainer for string formats).

PR #4464, opened 2022-02-05
fix: robotoff should not make the app crash if unavailable

### LINKED ISSUE #4357: App should not crash if Robotoff is down
### What

- Today robotoff was down (502 from nginx). As a result the android app in last versions was crashing just after product scan. https://sentry.io/organizations/openfoodfacts/issues/2653173009/
- It did not crash on older version (eg. v 433).

### Reproduction steps

1. Shut down robotoff
2. Scan a product
3. Just after scan happens, info on product appears for a short time, then app crash and you are redirected to welcome screen.

### Expected behavior

Android app should not crash, but just handle the 502 silently, not displaying any robotoff question.

VaiTon:
- Closes https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/4357


teolemon (approved):

R4md4c (approved):
LGTM 🚀

PR #4496, opened 2022-02-20
fix: Attempt to fix NPE: interface method 'java.lang.Object java.util.Map.get(java.lang.Object)

R4md4c:
####  Description
Unfortunately, I wasn't able to reproduce the crash. I tried to ensure that `productDetails` is not ever to be null which can happen when no-arg Constructor is called during de-serialization by the system. I'd suggest trying out this change and seeing if the issues disappear :crossed_fingers: 

Also, I've fixed a bug where I noticed the conversion from database value to a Map, were never correct and an empty hashmap is always returned when trying to read from DB. I've fixed that as well. 

#### Related issues
Hopefully a fix for #4479


teolemon:
@R4md4c @VaiTon merging and deploying to internal

R4md4c:
@teolemon Please let me know how it goes, otherwise we need to look for a different approach.

PR #4515, opened 2022-03-02
build: run unit tests and build apks in the same job during CI

VaiTon:


VaiTon:
Apparently, it runs better when parallelized. Closing for now.

PR #4531, opened 2022-03-13
fix: Sign in button is disabled outside the updateLoginButtonState method

### LINKED ISSUE #4530: Login screen: the sign-in button doesn't work after a failed attempt
### What
If a user inputs a wrong password in the sign-in screen, the button doesn't do anything after an edit to the username and/or password

1. Open the sign-in screen
2. Input a username and a wrong password
3. Click on the Sign-in button
4. An error message is displayed
5. Input another password
6. The button seems to be disabled

### Expected behavior
The button should be clickable after a failed attempt

### Important information
 - Device: Google Pixel 6
 - OS: Android 12
 - App Version: 3.6.8

g123k:
Fix for issue #4530

The sign-in button is disabled manually before all requests and never re-enabled after.
This is useless since updateLoginButtonState already handles all the different cases (loading, error…)


VaiTon (approved):

VaiTon:
Thanks for the PR! 🎉

Issue #4534, opened 2022-03-14
The flash button toggle doesn't work when the Edittext is visible


g123k:
### What
In the barcode scanner, if the text input dialog is visible, the flash button is clickable but doesn't do anything

### Steps to reproduce the behavior
1. Open the barcode scanner
2. Enter in "text mode"
3. Click on the flash toggle

### Expected behavior
Two options here: hide the button or make it work 😋

g123k:
I can work on a fix for this issue if you want.

But what would be the right way to fix it? Hide the button or make it work?
Cc @teolemon or @VaiTo

VaiTon:
I would prefer if it was hidden while in "input barcode mode", as the user is now focusing on another part of the UI. Any opinion @teolemon ?

g123k:
Thanks for your reviews @teolemon & @VaiTon 
It's now fixed

PR #4538, opened 2022-03-19
fix: Flash toggle should not be displayed when the quick view is visible

g123k:
Fix for issue #4534 

I have also added that if the quick view is visible, the back button should close it


VaiTon (approved):
LGTM! ✅ 

teolemon:
merging and releasing @VaiTon @g123k 
I created #4542 based on @VaiTon 's comment

PR #4549, opened 2022-03-19
fix: Vector drawables are not supported with drawableStart on Android < 5

g123k:
Fix for issue https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/4474

On Android < 5,  drawables passed to a drawableStart attribute (in XML) can not be a vector 
But doing it in the code is actually possible


g123k:
Sorry wrong branch used

PR #4553, opened 2022-03-19
fix: Sorting history items should not be done in the main thread

### LINKED ISSUE #4494: Hangup if you have a very long scan History (while refreshing the scores)
### What
- Very long histories hang while refreshing the scores
- Is there a way to make it non blocking ?

g123k:
I'm not sure if this will fix #4494, but right now, sorting items in the history list is done in the main thread.


teolemon (approved):

teolemon:
Thanks @g123k , merging

PR #4560, opened 2022-03-20
feat: The Activity now notifies when the drawer status has changed

### LINKED ISSUE #4540: UX problems on the search by text screen
Hello,

All the following issues are detected on the search by text screen:

- [ ] The text field should obtain the focus when the screen is visible (#4560)
- [x] The enter key of the keyboard should be a search icon (https://github.com/openfoodfacts/openfoodfacts-androidapp/pull/4541)
- [x] Taping on the enter key should launch a request (https://github.com/openfoodfacts/openfoodfacts-androidapp/pull/4541)

g123k:
To fix #4540 and more particularly the issue with the focus, I had to make some changes.
Right now, the `MainActivity` has the following lifecycle:

- [User] Click to open the drawer
- [Activity] Close the keyboard
- [User] Select of choice
- [Activity] Change the fragment and close the keyboard

The problem here is that the last keyboard event happens after the fragment is added/attached (after the `onResume`).

To fix this, the Fragments now can listen to drawer events.
Here is the new lifecycle:

- [User] Click to open the drawer
- [Activity] Close the keyboard
- [User] Select of choice
- [Activity] Change the fragment, close the keyboard and notify the attached Fragment
- [Fragment] Can force the keyboard to be reopened

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/MainActivity.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/MainActivity.kt
@@ -164,7 +164,7 @@ class MainActivity : BaseActivity(), NavigationDrawerListener, NavigationDrawerH
     private var userSettingsURI: Uri? = null
 
     private var historySyncJob: Job? = null
-    private var drawerStatusChanged: OnNavigationDrawerStatusChanged? = null
+    private val drawerStatusChangedListeners = mutableSetOf<OnNavigationDrawerStatusChanged>()
 
     private val loginThenUpdate = registerForActivityResult(LoginContract())
     { isLoggedIn -> if (isLoggedIn) updateConnectedState() }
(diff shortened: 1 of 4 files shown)


teolemon:
@g123k small merge conflicts to solve

teolemon:
@g123k should I keep both imports ?

g123k:
> @g123k should I keep both imports ?

Yes you can

VaiTon:
@g123k we should specify that with this approach the activity can only have a fragment listening to the drawer at any given time. Maybe also implement a control on the `onAttach` to check if the listener was null before attaching.

g123k:
Thanks for your feedback @VaiTon, I will add them

g123k:
@VaiTon I've changed my implementation to allow multiple listeners (with a `Set`).
Is-it better with this approach?

VaiTon (approved):

VaiTon:
Sorry for the delay!

Issue #4558, opened 2022-03-19
Invisible Bottom bar 


g123k:
### What
If the search panel is opened and then the keyboard is closed by the user, the bottom bar is no more visible.

### Steps to reproduce
- Launch the app
- Click on the search button in the Toolbar
- Close the keyboard

### Important information
 - Device: Pixel 6 Pro
 - OS: Android 12
 - App Version: 3.6.6

![Screenshot_20220320-001247.png](https://user-images.githubusercontent.com/246838/159141553-5d401de5-283a-4f2c-9ef9-7c5924705ae4.png)

g123k:
Fixed by #4574

PR #4574, opened 2022-03-21
fix: SearchView : ensure that the bottom navigation bar is hidden when the keyboard is visible and vice-versa

g123k:
A fix for #4558

This should fix all issues we have with the keyboard and the SearchView in the MainActivity
For your knowledge: `setOnActionExpandListener` is never called


teolemon (approved):

teolemon:
thanks @g123k 🌮

PR #4575, opened 2022-03-21
fix: In some cases, the user password is not sent to the API

### LINKED ISSUE #3520: Image upload is not properly credited to the contributor
### What
Image upload is not properly credited to the contributor

### Screenshot
image

### Part of
- #4202

g123k:
The user password is not always sent to the API.
This should resolve #3520


teolemon (approved):

VaiTon:
We should implement something to take care of that `getString("user"...` which is horrible and used in all the code :)

PR #4579, opened 2022-03-21
fix: Nutrition edition: no listener was set on unit/"data per"

### LINKED ISSUE #4470: [Bug]: Application does not take into account "per portion" switch when entering nutrition facts
### Open Food Facts username

postmalight

### What happened?

I report it on behalf of the user:

> When I try to enter the nutritional value of foods per serving, the application automatically puts me by 100ml/100g, but keeping the values entered for a smaller portion.
> I did another test with another food, I immediately chose the portion option before entering the information. Once saved, when I go back to the nutrition tab the portion is 100g.

> I've attached screenshots to show what's going on.
> Bug 2
> Bug 1

> Also I think it may be due to some information already entered by some users before me, but who did not specify the portion and which therefore automatically goes to 100 g

1. Scan a product without nutrition info
2. Edit nutrition info
3. Add a portion size in g
4. select "per portion" instead of per 100g
5. add values and validate
6. the data for a portion is shown as "per 100g"

### What device were you using ?

I'm not sure of Android Version. I don't know the phone

### Android Version

Android 12

### Barcode(s) of the product(s)

[0057700215036](https://fr.openfoodfacts.org/produit/0057700215036/chargement-maynards)

### Language your app was in

French

### Version

3.6.5 (Default)

### Flavor

Open Food Facts

### Installation method

Google PlayStore

### Relevant log output

_No response_

g123k:
I think I've found why the "nutrition_data_per" was always wrong: no listener was set on the radio group.
This should resolve #4470

(But please test it, as I'm not 100% sure)

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/edit/nutrition/ProductEditNutritionFactsFragment.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/edit/nutrition/ProductEditNutritionFactsFragment.kt
@@ -121,7 +121,12 @@ class ProductEditNutritionFactsFragment : ProductEditFragment() {
         binding.btnAddImageNutritionFacts.setOnClickListener { addNutritionFactsImage() }
         binding.btnEditImageNutritionFacts.setOnClickListener { newNutritionFactsImage() }
         binding.btnAdd.setOnClickListener { next() }
-        binding.for100g100ml.setOnClickListener { checkAllValues() }
+        binding.radioGroup.setOnCheckedChangeListener { _, checkedId ->
+            if (checkedId >= 0) {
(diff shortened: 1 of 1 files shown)


teolemon:
I'm going to merge and deploy it to internal so that we can thoroughly test.

teolemon:
- This now seems to work. In the process, I uncovered #4582 
image

g123k:
I have noticed this behavior yesterday, but wasn't sure if it was a bug or not.
If the switch is set to ON, we should hide the rest of the UI? And the same goes for the data sent to the server, isn't it?

teolemon:
We should hide it, but reshow it when we untick the toggle, and ensure the boolean is sent each time.

g123k:
Ok I will fix that

PR #4583, opened 2022-03-22
fix: "No nutrition facts" switch not working as intended

### LINKED ISSUE #4582: The "No Nutrition facts" toggle does not work correctly
### What
- The "No Nutrition facts" toggle does not work correctly

### Part of
- #4166 

### Video
https://user-images.githubusercontent.com/1689815/159445100-18216220-e401-427e-bb03-4d3ffc579b03.mp4

g123k:
PR that fixes two issues:
- If Nutrition data is set to OFF, the table beneath shouldn't be displayed
- If Nutrition data is set to OFF, no nutritional data is sent to the API

 #4582

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/edit/nutrition/ProductEditNutritionFactsFragment.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/edit/nutrition/ProductEditNutritionFactsFragment.kt
@@ -127,6 +127,10 @@ class ProductEditNutritionFactsFragment : ProductEditFragment() {
             }
         }
 
+        binding.checkboxNoNutritionData.setOnCheckedChangeListener { _, isChecked ->
+            binding.nutritionFactsLayout.visibility = if (isChecked) View.GONE else View.VISIBLE
+        }
+
         binding.btnAddANutrient.setOnClickListener { displayAddNutrientDialog() }
 
         binding.salt.doAfterTextChanged { updateSodiumValue() }
@@ -551,9 +555,10 @@ class ProductEditNutritionFactsFragment : ProductEditFragment() {
         // Add no nutrition data entry to map
         if (binding.checkboxNoNutritionData.isChecked) {
             targetMap[ApiFields.Keys.NO_NUTRITION_DATA] = "on"
-        } else {
-            targetMap += getNutrientsModeMap()
+            return targetMap
         }
+
+        targetMap += getNutrientsModeMap()
 
         // Add serving size entry to map if it has been changed
         if (binding.servingSize.isNotEmpty()) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/edit/nutrition/ProductEditNutritionFactsFragment.kt
```diff
@@ -551,10 +555,11 @@ class ProductEditNutritionFactsFragment : ProductEditFragment() {
         // Add no nutrition data entry to map
         if (binding.checkboxNoNutritionData.isChecked) {
             targetMap[ApiFields.Keys.NO_NUTRITION_DATA] = "on"
```
teolemon: So we send the boolean to deactivate, but do we send the boolean to reactivate, along or before nutrients ?

g123k: You're right, this is not the case right now.

Currently in the code, the app checks if the field is null (= no nutrition data) or == "ON"
So I will just send a null in the other cases. Otherwise I am afraid of generating some bugs

g123k: That should be OK right now, with the new commit [#9ce02dd](https://github.com/openfoodfacts/openfoodfacts-androidapp/pull/4583/commits/9ce02dd82ef53fbaf27f6126513ca9141e652a7c)

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/edit/nutrition/ProductEditNutritionFactsFragment.kt
```diff
@@ -575,6 +575,10 @@ class ProductEditNutritionFactsFragment : ProductEditFragment() {
             targetMap += getNutrientMapIfUpdated(view)
         }
 
+        // We force a null value here for "no nutrition", instead of an "off" String,
+        // to be sure to do not break other apps
+        targetMap[ApiFields.Keys.NO_NUTRITION_DATA] = null
```
teolemon: ?

teolemon: @stephanegigandet what are the expected values ? I'm opening some documentation issue on this

teolemon: @g123k : https://github.com/openfoodfacts/api-documentation/issues/70

g123k: Look at line 335, on Android the check is if != null -> true

I have improved the checks to ensure, "on" is true and everything else is false

g123k:
The code should now be OK with comments from https://github.com/openfoodfacts/api-documentation/issues/70

teolemon:
thank you @g123k , merging and releasing to internal 👍

PR #4586, opened 2022-03-22
ci: Ensure Labeler is not triggered on forks

teolemon:
####  Description
ci: Ensure Labeler is not triggered on forks since otherwise the test will fail with integration not accessible


g123k:
LGTM (I can't merge it, but you should @teolemon)

PR #4587, opened 2022-03-22
fix: Empty search list after a research by barcode

### LINKED ISSUE #4552: Empty search screen
### What
- After a search by barcode, the list with results is shown with an infinite loader

1. On the main screen, enter a barcode (eg: 3222476547255) and start the search
2. The product page will be shown
3. Press the back button
4. The products list search screen is shown with an infinite loader

https://user-images.githubusercontent.com/246838/159132581-b9e19b18-e521-4e02-aaba-412764cad7e7.mp4

g123k:
Will fix #4552

When a barcode is searched directly through the SearchView, the Search Activity is shown briefly… and then the Product Activity. But if we press the back button from the product, we will see the search activity with an infinite loader.
This page should have been finished before, which is now possible thanks to a dedicated listener

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -70,12 +70,7 @@
             android:name="androidx.startup.InitializationProvider"
             android:authorities="${applicationId}.androidx-startup"
             android:exported="false"
-            tools:node="merge">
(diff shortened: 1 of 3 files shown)


teolemon:
thanks for the attention to detail 👌  @g123k merging and deploying

PR #4608, opened 2022-03-25
fix: Useless swipe to refresh with an empty product

### LINKED ISSUE #4606: Product not found: infinite loader
- Open an unknown product
- Swipe to refresh
- The progress bar will stay forever
- 
a

g123k:
If a product doesn't exist, the app shows a dedicated screen to create it.
On this screen, a swipe to refresh is available: but to refresh what?

That's why this PR disables the swipe to refresh


teolemon:
This view normally holds search results that you can refresh @g123k 
EDIT: I read too quickly, LGTM

teolemon:
Ah the other case could be server issues.

teolemon:
Actually, could we jump directly to the product addition screen ?

teolemon (approved):

g123k:
You're right, it would be better to open the edition screen instead.
I change that

g123k:
Mmmm actually I think we should stay on the current implementation (= this PR is still relevant).

How it works right now:
- If this is a valid barcode -> create a new product
- If this is an invalid barcode (the case here) -> error message asking to take a picture -> create a new product

teolemon:
Merged as a result

Issue #4613, opened 2022-03-26
Small darkmode error in the Contributions tab


teolemon:
### What
- Small darkmode error in the Contributions tab

### Screenshot
![image](https://user-images.githubusercontent.com/1689815/160239321-b1556cac-4eb3-4584-8b40-9c7906c92ba6.png)

galemu2:
I found the bug.  The `TextView` has a text color value that set set to black, `android:textColor="@color/black"`.  Possible solutions, are remove the text color. This view will use materials color.  Define text color in `colors.xml` file in both `colors` and `colors-night` folder. Or. use `android:textAppearance` with custom *style* or one inherited from Android platform's default text appearance. 

Since the `attribute_group_item.xml` file uses the `textAttribute` inplace of `textColor`, I think the `textAttribute` solution, using *inherited* **style** from Android platform sounds appropriate.  `android:textAppearance="@android:style/TextAppearance.DeviceDefault"`

VaiTon:
Fixed by https://github.com/openfoodfacts/openfoodfacts-androidapp/pull/4628

PR #4628, opened 2022-04-05
Fix Issue #4613: Dark mode error in Contributions tab

galemu2:
`TextView` In fragment_contributors.xml the TextView for "Product
States" has text color set to black, which kept the text black when in
dark mode. After fix, the text color is light when in dark mode, and
dark when dark in light mode.

I used `textAppearance` because it is used in `attribute_group_item.xml`


VaiTon (approved):

teolemon:
Thank you and congratulations @galemu2 for your PR to @openfoodfacts 👍 merging

PR #4633, opened 2022-04-07
refactor: extract abstract class CameraView

### LINKED ISSUE #4542: Create an abstraction for the scanner

_Originally posted by @VaiTon in https://github.com/openfoodfacts/openfoodfacts-androidapp/pull/4538#pullrequestreview-915003230_

VaiTon:
Closes https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/4542


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/scan/ContinuousScanActivity.kt
```diff
@@ -512,31 +501,19 @@ class ContinuousScanActivity : BaseActivity(), IProductView {
...
-            binding.barcodeScanner.barcodeView.decoderFactory = DefaultDecoderFactory(ScannerPreferencesRepository.BARCODE_FORMATS)
-            binding.barcodeScanner.setStatusText(null)
-            binding.barcodeScanner.setOnClickListener {
-                quickViewBehavior.state = BottomSheetBehavior.STATE_HIDDEN
-            }
-            binding.barcodeScanner.barcodeView.cameraSettings.run {
-                requestedCameraId = cameraState
-                isAutoFocusEnabled = autoFocusActive
-            }
+        val useMLScanner = BuildConfig.USE_MLKIT &&
```
g123k: This value is also obtained in ScannerPreferencesRepository.
Maybe a single method for this file & ScannerPreferencesRepository would be better

g123k (changes requested):

app/src/main/res/layout/activity_continuous_scan.xml
```diff
@@ -33,12 +33,14 @@
         <com.journeyapps.barcodescanner.DecoratedBarcodeView
```
g123k: Wouldn't it be better to have the ZXing implementation also in a ViewStub?
Because if I understand well, when we use ML Kit, this view will be created (even if gone)

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/scan/ZXCameraView.kt
```diff
@@ -0,0 +1,96 @@
...
+            override fun possibleResultPoints(resultPoints: MutableList<ResultPoint>?) = Unit
+        })
+    }
+
+    override fun onResume() {
+        decoratedView.resume()
+    }
+
+    override fun updateFlashSetting(flashActive: Boolean) {
+        if (flashActive) decoratedView.setTorchOn()
```
g123k: Please be consistent with brackets.
On line 57, you use them and not here ;)

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/scan/ContinuousScanActivity.kt
```diff
@@ -506,37 +494,23 @@ class ContinuousScanActivity : BaseActivity(), IProductView {
...
-            }
-            binding.barcodeScanner.barcodeView.cameraSettings.run {
-                requestedCameraId = cameraState
-                isAutoFocusEnabled = autoFocusActive
-            }
 
-            // Start continuous scanner
-            binding.barcodeScanner.decodeContinuous(barcodeScanCallback)
-            beepManager = BeepManager(this)
+        cameraView = if (!scannerPrefsRepository.mlScannerEnabled) {
```
g123k: (To discuss) Wouldn't it be better to have a CameraView factory somewhere else?
So we can use it in multiple screens (just in case)

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/scan/ContinuousScanActivity.kt
```diff
@@ -804,31 +721,26 @@ class ContinuousScanActivity : BaseActivity(), IProductView {
...
-            lastBarcode = result.text
-            if (!isFinishing) {
-                setShownProduct(result.text)
-                matomoAnalytics.trackEvent(AnalyticsEvent.ScannedBarcode(result.text))
-            }
+        val invalidBarcode = daoSession.invalidBarcodeDao.unique {
+            where(InvalidBarcodeDao.Properties.Barcode.eq(text))
         }
+        // Scanned barcode is in the list of invalid barcodes, do nothing
+        if (invalidBarcode != null) return
```
g123k, line 734: Missing brackets here :)

g123k (changes requested):
Just a missing bracket ✊

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/scan/ContinuousScanActivity.kt
VaiTon: I usually don't put brackets if I only return. We should set a checkstyle here :)

g123k:
Thanks for your awesome @VaiTon, I merge your work 👍

g123k (approved):

g123k:
Ooops it seems some tests are failing now.
We should check them out

g123k:
@VaiTon I have to revert this PR, as your code doesn't compile
https://github.com/openfoodfacts/openfoodfacts-androidapp/runs/5875416786?check_suite_focus=true#step:5:1212

Issue #4640, opened 2022-04-07
Add new matomo reporting 


teolemon:
### What
- Add new matomo reporting 
  - # of products in history
  - # of products added to lists
  - each time a product is added to a list ?
- That should enable us to assess criticality of data migration

g123k:
Shouldn't we also monitor how many lists / products a user have?
Because you will only monitor changes with this kind of reporting

g123k:
#4648 will trigger an event when a product is added or removed from a list

PR #4648, opened 2022-04-08
fix: wrong number of products for lists + analytics events

### LINKED ISSUE #4647: Incorrect number of products in a list
### What
After a product added to a list, the "number of products" counter is invalid

(optional, delete if empty)
1. Open the Lists tab
2. Go back to the homescreen and search for a product
3. Select a product
4. Click on "Add to list"
5. Go back to the Lists tab
6. The number of elements in the list is incorrect

### Screenshot/Mockup/Before-After
https://user-images.githubusercontent.com/246838/162414168-a23cabbd-08d0-4fa7-8faf-731fc9cce8f1.mp4

g123k:
Fix for #4647 and will partially implement #4640

ProductsListsDao was never updated after a product addition/removal
+ analytics events when a product is added / removed
+ analytics event when a list is shared

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/analytics/AnalyticsEvent.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/analytics/AnalyticsEvent.kt
@@ -26,7 +26,7 @@ sealed class AnalyticsEvent(val category: String, val action: String, val name: 
 
     object RobotoffLoggedInAfterPrompt : AnalyticsEvent("user-account", "logged-in-after-prompt", "robotoff", null)
 
-    data class ShoppingListCreated(val listName: String?) : AnalyticsEvent("shopping-lists", "created", listName, null)
+    object ShoppingListCreated : AnalyticsEvent("shopping-lists", "created", null, null)
 
     data class ShoppingListProductAdded(val barcode: String) : AnalyticsEvent("shopping-lists", "add_product", barcode, null)
 
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/analytics/AnalyticsEvent.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/analytics/AnalyticsEvent.kt
@@ -26,7 +26,13 @@ sealed class AnalyticsEvent(val category: String, val action: String, val name: 
 
     object RobotoffLoggedInAfterPrompt : AnalyticsEvent("user-account", "logged-in-after-prompt", "robotoff", null)
 
-    object ShoppingListCreated : AnalyticsEvent("shopping-lists", "created", null, null)
+    data class ShoppingListCreated(val listName: String?) : AnalyticsEvent("shopping-lists", "created", listName, null)
+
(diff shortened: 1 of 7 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/analytics/AnalyticsEvent.kt
```diff
@@ -26,7 +26,13 @@ sealed class AnalyticsEvent(val category: String, val action: String, val name:
 
     object RobotoffLoggedInAfterPrompt : AnalyticsEvent("user-account", "logged-in-after-prompt", "robotoff", null)
 
-    object ShoppingListCreated : AnalyticsEvent("shopping-lists", "created", null, null)
+    data class ShoppingListCreated(val listName: String?) : AnalyticsEvent("shopping-lists", "created", listName, null)
```
VaiTon: Do we really want to know about the product list name? Isn't it something a little bit privacy-concerning?

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/summary/SummaryProductFragment.kt
```diff
@@ -884,6 +884,11 @@ class SummaryProductFragment : BaseFragment(), ISummaryProductPresenter.View {
                 it.imageUrl = product.getImageSmallUrl(localeManager.getLanguage())
             }
             daoSession.listedProductDao.insertOrReplace(product)
+            daoSession.productListsDao.update(list.apply {
+                products.add(product)
+                numOfProducts++ }
+            )
+            matomoAnalytics.trackEvent(AnalyticsEvent.ShoppingListProductAdded(product.barcode))
             dialog.dismiss()
             onRefresh()
```
VaiTon, line 890: Can we extract this to a fun?

app/src/main/java/openfoodfacts/github/scrachx/openfood/analytics/AnalyticsEvent.kt
g123k: My idea behind thisn was to differentiate what kind of list the user create.
But clearly I can remove it if neccessary. What do you think @teolemon?

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/summary/SummaryProductFragment.kt
g123k: No pb

g123k:
That should be ok now @VaiTon

app/src/main/java/openfoodfacts/github/scrachx/openfood/analytics/AnalyticsEvent.kt
teolemon: probably overkill. We're more focused on volumetry.

g123k:
Following @teolemon answer, no list name is sent with the list creation analytics event

VaiTon (approved):

PR #4656, opened 2022-04-09
refactor: extract CameraView interface

### LINKED ISSUE #4542: Create an abstraction for the scanner

VaiTon:


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/scan/CameraView.kt
```diff
@@ -0,0 +1,49 @@
...
+import android.widget.FrameLayout
+import androidx.annotation.CallSuper
+import androidx.annotation.LayoutRes
+import androidx.appcompat.app.AppCompatActivity
+import androidx.core.view.isVisible
+
+abstract class CameraView<T : FrameLayout>(
+    protected val activity: AppCompatActivity,
+    protected val viewStub: ViewStub,
+    @LayoutRes private val resId: Int,
```
g123k: A nullable value could be better here, instead of 0

app/src/main/java/openfoodfacts/github/scrachx/openfood/features/scan/ZXCameraView.kt
```diff
@@ -0,0 +1,107 @@
...
+    R.layout.view_camera_zx_preview
+) {
+    private val beepManager by lazy { BeepManager(activity) }
+
+    override fun attach(cameraState: Int, flashActive: Boolean, autoFocusActive: Boolean) {
+        super.attach(cameraState, flashActive, autoFocusActive)
+
+        view.apply {
+            isVisible = true
+            visibility = View.VISIBLE
```
g123k: What's the difference with the previous call?

g123k (changes requested):

g123k (approved):

g123k:
Perfect, thanks a lot @VaiTon for your PR 👍

PR #4657, opened 2022-04-09
fix: When nutriscore is not (yet) computed, hide the "Learn more" button

### LINKED ISSUE #4620: Only links and tappable words should be written in blue
### Describe the bug

There are 3 subheadings, all in blue text but one of them isn't tappable.
There should be a difference in text color between words that are tappable and those that aren't

### To Reproduce

open food facts exercise
nutri-score

### Expected behavior

I expected the blue link to be tappable either as a dropdown or leading to another page, else it should be in black like every other normal text

### Screenshots

Screenshot_20220328-164820
The blue text, "Learn more about the nutri-score", doesn't open or lead anyway. It looks like it should be tappable but it isn't. I kept tapping thinking the issue is with my phone screen:)

### Additional context

_No response_

### Type of device

Other

### Browser version

_No response_

### Number of products impacted

Products affected are ones without a nutri-score calculated yet

### Time per product

Time loss: 30secs

g123k:
Will fix #4620

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/nutrition/NutritionProductFragment.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/nutrition/NutritionProductFragment.kt
@@ -220,6 +220,8 @@ class NutritionProductFragment : BaseFragment(), CustomTabActivityHelper.Connect
             }
         }
 
+        binding.nutriscoreLink.visibility = if (product.nutritionGradeFr != null) VISIBLE else GONE
+
         var servingSize: Measurement? = null
         val servingSizeString = product.servingSize
         if (servingSizeString.isNullOrEmpty()) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/product/view/nutrition/NutritionProductFragment.kt
```diff
@@ -220,6 +220,8 @@ class NutritionProductFragment : BaseFragment(), CustomTabActivityHelper.Connect
             }
         }
 
+        binding.nutriscoreLink.visibility = if (product.nutritionGradeFr != null) VISIBLE else GONE
```
VaiTon:
```suggestion
        binding.nutriscoreLink.isVisible = product.nutritionGradeFr != null
```
From [core-ktx](https://android.github.io/android-ktx/core-ktx/androidx.view/android.view.-view/is-visible.html)

VaiTon (changes requested):

VaiTon (approved):

g123k:
@teolemon & @VaiTon If it's ok for both, could we merge this?

Issue #4600, opened 2022-03-23
Use the Google Play "In-App-Review" API


g123k:
Instead of opening Google Play to rate the app, we should use the embedded API "In-App-Review":
https://developer.android.com/guide/playcore/in-app-review

![iar-flow](https://user-images.githubusercontent.com/246838/159771550-94d6cbf8-3a17-4b55-8d2f-f12cc0f49bc4.jpeg)

Note: Only Google Play variants should implement this

lailajs:
could I tackle this? Thanks!

g123k:
My initial idea was to let my students implement it tomorrow.
it does not bother you @lailajs ?

g123k:
You can take it @lailajs (my students won't have the time to work on it today)

g123k:
Is-it ok for you @lailajs?

g123k:
Any news @lailajs?

lailajs:
Yes! I’ll be working on it this week as part of a school project. Thank you

lailajs:
hey @g123k, im looking through the repo code, and was wondering if you have any ideas of where the best  place to implement this feature is. Thank you so much!

g123k:
It's already implemented in the app.
In the MainActivity, you have the `showFeedbackDialog` method, which is called on a user is logged in

lailajs:
ohh okay I see! So what exactly is my task then? the current code makes the user leave the app right?

g123k:
The current code doesn’t use the Rating API from the Google Play.
That’s the change required in this issue

lailajs:
Yes, I see. Just to make sure, the current code redirects the user to the google play store, right? Im a beginner in Kotlin and just want to make sure!

g123k:
Yes.
You should test the app, you will how the workflow works

lailajs:
Do you know if its possible to run the android version on my iphone? i dont have any android devices, and I'm not sure how to run the app without one

g123k:
There is an emulator on Android, directly available on Android Studio

lailajs:
its so weird, when i run the app it only runs open beauty facts.. not open food facts, does that happen for you?

g123k:
I don't want to offense you, but regarding your questions I suggest you to take some time to learn the basics of the Android programming

PR #4690, opened 2022-04-19
feat: Added In-App Review Functionality - Fixing Issue #4600 

lailajs:
####  Description
Hello! I added the functionality of the Google In-App Review Functionality, as labeled in issue 4600. Thank you!

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -33,6 +33,7 @@ fun obtainTestBuildType(): String {
     return project.properties.getOrDefault("testBuildType", result) as String
 }
 
+
 dependencies {
     // Kotlin
     implementation("org.jetbrains.kotlin:kotlin-stdlib:${rootProject.extra["kotlinVersion"]}")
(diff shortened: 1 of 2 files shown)


g123k (changes requested):
Thanks for your PR @lailajs, but your implementation is available for all flavors.
As we also publish to Fdroid, this API should only be used with playStore flavor.

Could you make that change please?

lailajs:
> Thanks for your PR @lailajs, but your implementation is available for all flavors. As we also publish to Fdroid, this API should only be used with playStore flavor.
> 
> Could you make that change please?

Hi! I am not exactly sure what that means, could you help me understand it better? Thanks!

VaiTon:
> > Thanks for your PR @lailajs, but your implementation is available for all flavors. As we also publish to Fdroid, this API should only be used with playStore flavor.
> > Could you make that change please?
> 
> Hi! I am not exactly sure what that means, could you help me understand it better? Thanks!

We publish two version of the app with little modifications. The first is for the Google Play Store, the second is for FDroid. For the FDroid users we do not include google libraries only available to those who installed the playstore as it is often used on OS without it.

You can find the definition of the two flavors in the app/build.gradle.kts. You should put the code only in the fdroid source folder and the fdroid flavor.

VaiTon:
Superseded by https://github.com/openfoodfacts/openfoodfacts-androidapp/pull/4860

PR #4777, opened 2022-06-22
build: adopt Gradle Version Catalog

chris-hatton:
####  💬 Description
This PR adopts Gradle's [Version Catalogs](https://docs.gradle.org/current/userguide/platforms.html) feature for dependency management (considered stable from version `7.4`).
Adoption means moving all dependency coordinates and version numbers to a newly created `libs.versions.toml` file.
This becomes referenced by the existing Gradle build files.

To accompany this change, I have also integrated two plugins that assist in updating the Version Catalog:
- Ben Manes' popular [Gradle Versions Plugin](https://github.com/ben-manes/gradle-versions-plugin) which can check for new versions of dependencies.
    - This has been configured to update only to versions that don't include _'Alpha'_, _'Beta'_, _'RC'_ or _'EAP'_
    i.e. only stable versions. 
- LittleRobots [Version Catalog Update Plugin](https://github.com/littlerobots/version-catalog-update-plugin) uses previous plugin to re-write the `libs.versions.toml` file on demand, updating all new dependencies.

#### ⭐ Motivation

My motivation for wanting the OpenFoodFacts App to adopt **Gradle Version Catalogs** arises from wanting **a Kotlin Client Library for accessing OpenFoodFacts API's**.
As far as I'm [aware](https://openfoodfacts.slack.com/archives/C02KVRT2C/p1643630052370229), a fully featured Kotlin API Client only _really_ exists today as part of the Android App.

_How is this desire related to Version Catalogs?_

I plan to undertake the work to **extract the API portion of the _Open Food Facts_ Android Client App** to a separate project repository and Gradle-build, to be re-included as a [composite Gradle build](https://docs.gradle.org/current/userguide/composite_builds.html).  Having the dependency versions extracted as a **Version Catalog** enables easier sharing of versions between distinct Gradle project and builds.

Overall, moving to **Version Catalog**s represents both a modernisation of the projects dependency management but also an enabler to extract a **Kotlin API Client** that can be concurrently maintained with the Mobile client.

#### ⚠️ Impact on version updates

- Currently, dependency update PR's are being [automatically raised](https://github.com/openfoodfacts/openfoodfacts-androidapp/pulls/app%2Fdependabot) against this Android App project by [Dependabot](https://github.com/dependabot).  This bot is not _currently_ aware of Version Catalogs and so would stop automatically raising PR's.  There is an [open issue](https://github.com/dependabot/dependabot-core/issues/3121) with Dependabot to implement Version Catalog support.
- **After this change; dependency updates will be semi-automatic**: A project maintainer should periodically run the `versionCatalogUpdate` Gradle task which will cause the `libs.versions.toml` file to be re-written with new versions.  A PR can then be raised.
- This has both pro's and con's: On one hand a maintainer should remain aware of the task and won't be prompted by Dependabot. On the other hand, being able to manually review bulk changes to versions may result in fewer undesired 'single update' PR's which I notice are creating a lot of 'noise' in the PR queue and burdensome to manage in their own right.

#### 🧑‍💼 Follow-on work

- After this PR is merged I will proceed with extracting the API Client portion of the App to a separate, composite module.  If this is successful I will offer from Open Food Facts maintainers to host the extracted Kotlin API Client Library as a new repository in the Open Food Facts organisation.

#### Related PRs
Blocked on https://github.com/openfoodfacts/openfoodfacts-androidapp/pull/4794


app/build.gradle.kts
```diff
@@ -15,13 +15,13 @@
  */
 plugins {
     id("com.android.application")
-    id("de.timfreiheit.resourceplaceholders.plugin")
+    alias(libs.plugins.resourceplaceholders) apply true
```
VaiTon, line 18: Why we do this on resourceplaceholders and not on the lines next to this one? I'm not an expert on GVC so just asking

gradle/libs.versions.toml
```diff
@@ -0,0 +1,233 @@
...
+    "jackson-annotations",
+    "jackson-core",
+    "jackson-databind",
+    "jackson-kotlin",
+]
+crash-analytics = [
+    "matomo",
+    "sentry-android",
+]
+unit-testing = [
```
VaiTon: I would rename it `testing` as we're also using libs for instrument tests
```suggestion
testing = [
```

settings.gradle.kts
```diff
@@ -1 +1,17 @@
 include(":app")
+
+dependencyResolutionManagement {
```
VaiTon, line 3: Don't we do this already in the root `build.gradle.kts`?

VaiTon (commented):
LGTM! Just a few considerations that I'd like to address before merging.

app/build.gradle.kts
chris-hatton: The need to do this for _some_ plugins relates to using the [Gradle plugins DSL](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block), rather than being specific to Version Catalogs. GVC encourages migrating all plugins to the `plugins` DSL block, which is why this has arisen now.

Some plugins do not manage their class-path/class loading order in quite the way that the newer plugins DSL expects. For such plugins it can be necessary to first load (but **not** apply) the plugin in the root Gradle project, bringing it into the class-path before the `app` module is loaded, and then continue to actually apply it in the `app` (or other) module as required.

Note that in the root project `plugins` block I added:
```
alias(libs.plugins.resourceplaceholders) apply false // Android Extension in `app` breaks without this
```

gradle/libs.versions.toml
chris-hatton: Good call, I'll rename it.

settings.gradle.kts
chris-hatton: You're right, I can remove `dependencyResolutionManagement` block.

The `pluginManagement` block should remain, to resolve some [custom repositories for the Gradle plugins DSL](https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_resolution_rules).

chris-hatton: On inspection, decided to remove the repeated definitions in `build.gradle.kts` in favour of keeping single repository definitions here in `settings.gradle.kts`.  [This now seems preferred](https://docs.gradle.org/current/userguide/dependency_management.html#sub:centralized-repository-declaration).

chris-hatton:
@VaiTon Apologies, there's still a build regression here but I've run out of time tonight.  Hope to resolve to bring this PR out of draft tomorrrow. ✌️

VaiTon:
> @VaiTon Apologies, there's still a build regression here but I've run out of time tonight. Hope to resolve to bring this PR out of draft tomorrrow. ✌️

No worries!

chris-hatton:
Brief update; some unexpected and unknown interaction between Version Catalogs and the way dependencies are being resolved, has caused a duplicate class issue between the mikepenz libraries:
```
Duplicate class com.mikepenz.iconics.typeface.IIcon found in modules jetified-iconics-core-4.0.2-runtime (com.mikepenz:iconics-core:4.0.2) and jetified-iconics-typeface-api-5.1.1-runtime (com.mikepenz:iconics-typeface-api:5.1.1)
```
...I have double checked the `isTransitive` setting on `materialDrawer` which I imagine may have been used to work around a similar issue previously; but this isn't resolving it.  I will look into alternate ways of resolving the duplicate without changing the actual dependency versions (aiming to limit the scope of changes in the PR).

chris-hatton:
I think the duplicate class issues associated with this **Version Catalog** adoption are resolved by my latest commit.

However, I was prevented from fully testing the App, by encountering these existing issues with Barcode Scanning:

- [Report 1](https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/4685)
- [Report 2](https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/4686)

Both issues point to a regression error that has been present in the `develop` branch since April. It looks like this would prevent maintenance releases.  Is this alarming? Is the new Flutter App the focus for all future releases?

I'll see if I can resolve it, which will also enable proper testing of this PR to increase confidence before merging.

VaiTon:
Hey @chris-hatton!

> Both issues point to a regression error that has been present in the `develop` branch since April. It looks like this would prevent maintenance releases. Is this alarming? Is the new Flutter App the focus for all future releases?

Actually the new Flutter App is aimed at the every day user in opposition to this app that is aimed at power contributors. Because of the recent surge in exposure of the flutter app we're dedicating more resources to that but it doesn't mean that this one should be abandoned! 

At the moment I'm a little busy with studying but feel free to do PRs and I'll promptly review and merge them!

VaiTon (approved):

chris-hatton:
That's great context to have thanks @VaiTon .  I'll pursue this blocking defect, PR and then move onto my intended goal of extracing the API portion of the App to a Kotlin Multiplatform library (first within the project, then perhaps externalised via Gradle Composite builds).

chris-hatton:
@VaiTon PR re-based to latest & checks passing 👌

VaiTon:
Thank you for your work @chris-hatton !! :tada:

PR #4794, opened 2022-07-07
fix: Fix crashes by ensuring layout ID is set before inflating Camera View stub + raise AGP for Kotlin 1.7.0 compatibility

### LINKED ISSUE #4685: RuntimeException: Unable to start activity ComponentInfo{org.openfoodfacts.scanner/openfoodfacts.github.scrachx.ope...
Sentry Issue: [OPENFOODFACTS-ANDROID-5F9](https://sentry.io/organizations/openfoodfacts/issues/3182652156/?referrer=github_integration)

```
IllegalArgumentException: ViewStub must have a valid layoutResource
    at android.view.ViewStub.inflate(ViewStub.java:284)
    at android.view.ViewStub.setVisibility(ViewStub.java:235)
    at openfoodfacts.github.scrachx.openfood.features.scan.CameraView.attach(CameraView.kt:58)
    at openfoodfacts.github.scrachx.openfood.features.scan.ZXCameraView.attach(ZXCameraView.kt:26)
    at openfoodfacts.github.scrachx.openfood.features.scan.ContinuousScanActivity.onCreate(ContinuousScanActivity.kt:503)
...
(12 additional frame(s) were not displayed)

RuntimeException: Unable to start activity ComponentInfo{org.openfoodfacts.scanner/openfoodfacts.github.scrachx.openfood.features.scan.ContinuousScanActivity}: java.lang.IllegalArgumentException: ViewStub must have a valid layoutResource
    at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2416)
    at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2476)
    at android.app.ActivityThread.-wrap11(ActivityThread.java)
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1344)
    at android.os.Handler.dispatchMessage(Handler.java:102)
...
(5 additional frame(s) were not displayed)
```

chris-hatton:
####  Description
Code was attempting to set the visibility of the View Camera stub before its source layout XML id was set.
This is an error because setting a View visible causes `inflate()` to be run in case it is not already inflated.
With no XML id set; there is nothing to inflate and so subsequent code that assumes a view to be present fails. 

Simply switching around the order of operations here...

1. Setting the layout ID
2. Inflating
3. Setting visibility

... resolves the issue.

#### Related issues
Fixes #4685, #4686

#### Related PRs
Unblocks testing of #4777 

#### Screenshots
N/A

#### Link to the automatically generated build APK
N/A


chris-hatton:
The CI linter is failing this build with errors in the pattern:
```
Error: Calling new methods on older versions: Call requires API level 24 (current min is 21): `<some kotlin stdlib method>`
```
...given this error is unrelated to the changes in the PR; it seems likely this is caused by an environmental change.
Has the linter been updated or reconfigured independently of the source code?

A direct answer to this error seems to be raising the min API level to 24, but I'm not sure that's expected or desirable.  Other maintainers please advise.

chris-hatton:
I've added raising min API level to 24 for expedience. Apparently, combining API 21 with JDK 8 `stdlib` (as current config implies) is a genuine error and probably should have been addressed before.

If it's important to keep API 21 compatibility then I can edit this PR again, but we'd need to explicitly use `stdlib-jdk7` and probably revisit the use of some JDK 8 `stdlib` functions throughout the code.

VaiTon:
> I've added raising min API level to 24 for expedience. Apparently, combining API 21 with JDK 8 `stdlib` (as current config implies) is a genuine error and probably should have been addressed before.
> 
> If it's important to keep API 21 compatibility then I can edit this PR again, but we'd need to explicitly use `stdlib-jdk7` and probably revisit the use of some JDK 8 `stdlib` functions throughout the code.

I think that we'd like to keep API 21 as min for now. If you could do the changes described to keep the minSdk to 21 it would be great!

chris-hatton:
Thanks @VaiTon ; looking again at the errors, I discovered they actually stem from incompatibility between the linter in Android Gradle Plugin `7.0.4` and Kotlin `1.7`.  The linter thinks the source is trying to use JDK 8 Stream operators instead of those included in the Kotlin `stdlib`.  I've applied a fix to upgrade to the latest AGP `7.2.1`, which is probably advisable anyway.

NikNoori:
Any update on this issue? Thanks.

chris-hatton:
> Any update on this issue? Thanks.

The crash fix itself was straightforward.
The complication with having it pass in CI stems from bit-rot: The minimally fixed code wouldn't build in CI because of (apparently) some unrelated change in the build environment.  That can be resolved by upgrading to Gradle 7.2.1 which entailed at least one further breaking change in the project.  Still working on it, just time poor.

@VaiTon I have applied a fix for the most recent run failure: un-mergeable Android Manifest of `androidTestScreenshots` module due to incorrect package name.  It may work next time if you can kindly re-run it.

app/build.gradle.kts
```diff
@@ -218,7 +218,7 @@ android {
     defaultConfig {
         applicationId = "openfoodfacts.github.scrachx.openfood"
 
-        minSdk = 21
+        minSdk = 24
```
VaiTon: Is this mandatory for upgrading the AGP?

chris-hatton: It is not! Apologies, I had meant to revert this according to our previous discussion, it will be included in next commit to resolve some language translations linting errors.

VaiTon:
@chris-hatton I've reverted the minSdk change

PR #4825, opened 2022-08-18
chore: Improved "Add to Open Food Facts modal"

### LINKED ISSUE #2966: Add a label to the "Add to Open Food Facts modal"
image
The line looks like a separator. Add a field title to make it clear where to type

pratyaksh1610:
####  Description

 Added a placeholder indicating the field to "Add to Open Food Facts modal".

#### Related issues
Fixes #2966 

### Actual behaviour
final

### Expected behaviour
expected

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -569,6 +569,7 @@
 
     <!-- strings related to share images -->
     <string name="sorry_msg">Pictures without barcode cannot yet be added with the app. Please try the mobile web version.</string>
+    <string name="type_the_barcode_here_label" translatable="false">Type the barcode here</string>
     <string name="no_barcode">We have not detected any barcode</string>
     <string name="enter_barcode">Please type the barcode below</string>
     <string name="format_error">Wrong Barcode/QR format</string>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -569,6 +569,7 @@
 
     <!-- strings related to share images -->
     <string name="sorry_msg">Pictures without barcode cannot yet be added with the app. Please try the mobile web version.</string>
+    <string name="type_the_barcode_here_label" translatable="false">Type the barcode here</string>
```
VaiTon:
```suggestion
    <string name="type_the_barcode_here_label">Type the barcode here</string>
```

VaiTon (changes requested):

app/src/main/res/values/strings.xml
pratyaksh1610: @VaiTon Is there any reason of removing `translatable="false"`. , When do we need to make it `true` and `false` ?

pratyaksh1610: After doing some exploration , I found that , when we use `translatable="false"`, it means that it will not be translated irrespective of the language (in case of multiple languages), so it will remain same in all states.

VaiTon: Yeah, and that's not what we want in this case

pratyaksh1610:
@VaiTon PTAL .
Thanks.

VaiTon (approved):

pratyaksh1610:
@VaiTon I want to ask a question, How to approach a particular issue ? What is your take on this ?

VaiTon:
Which issue are you referring to? You can talk to me on slack if you want!

PR #4871, opened 2022-10-30
feat: Use new SplashScreen API on Android 12 and newer (API level 31 and up)

### LINKED ISSUE #4548: feat: Implement Android 12 new splashscreen API
### What
Since Android 12, a brand new [SplashScreen API](https://developer.android.com/guide/topics/ui/splash-screen) is available.
In this OS version, the app actually have two splash screens:

https://user-images.githubusercontent.com/246838/159131417-7402751d-9435-4719-bf32-ee11d86657c8.mp4

fogleins:
####  Description
This PR implements the new SplashScreen API on Android 12 and newer, provides backwards compatibility using the Splash Screen compat library (on API < 31, the old splash screen is shown). The implementation works both with light and dark themes.
Tested on API 29 (backwards compatible) and API 32 (uses the new API).

#### Related issues
Fixes #4548 

##### API >= 31:
image
image

##### API < 31:
image
image

#### Link to the automatically generated build APK
[debug.zip](https://github.com/openfoodfacts/openfoodfacts-androidapp/files/9896879/debug.zip)

### DIFF
--- a/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/splash/SplashActivity.kt
+++ b/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/splash/SplashActivity.kt
@@ -2,8 +2,11 @@ package openfoodfacts.github.scrachx.openfood.features.splash
 
 import android.annotation.SuppressLint
 import android.content.pm.ActivityInfo
+import android.os.Build
 import android.os.Bundle
 import androidx.activity.viewModels
+import androidx.core.splashscreen.SplashScreen
+import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
 import androidx.core.view.isVisible
 import androidx.lifecycle.lifecycleScope
 import com.google.android.material.snackbar.BaseTransientBottomBar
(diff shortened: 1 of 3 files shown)


app/src/main/java/openfoodfacts/github/scrachx/openfood/features/splash/SplashActivity.kt
```diff
@@ -2,8 +2,11 @@ package openfoodfacts.github.scrachx.openfood.features.splash
 
 import android.annotation.SuppressLint
 import android.content.pm.ActivityInfo
+import android.os.Build
 import android.os.Bundle
 import androidx.activity.viewModels
+import androidx.core.splashscreen.SplashScreen
+import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
```
VaiTon, line 9: I would remove this import and use SplashScreen.installSplashScreen(), as we use this function just one time

fogleins: Since `installSplashScreen()` is an extension function on the `Activity` class, I'm not sure it's possible to exclude this import statement. See https://kotlinlang.org/docs/extensions.html#scope-of-extensions
Take a look at `installSplashScreen()`'s signature (code from the `androidx.core.splashscreen` package):
```kotlin
public fun Activity.installSplashScreen(): SplashScreen {
    //
}
```
I might be wrong, however, if so, please tell me how could I resolve this issue. Thanks!

VaiTon: You're right. I reviewed the code in github so no different colors for extensions! Thank yoU!

VaiTon:
Thank you @fogleins!

PR #4892, opened 2022-12-16
Fix: Added Null Checks

### LINKED ISSUE #4883: ProductsAPITest > post product() FAILED     retrofit2.HttpException at KotlinExtensions.kt:53
### What
- ProductsAPITest > post product() FAILED
    retrofit2.HttpException at KotlinExtensions.kt:53
- Latest release build failing

PeterHe0226:
####  Description
Added Null checks for Product and ProductState

#### Related issues
"Fixes #4883"


VaiTon:
Hey @PeterHe0226! Thank you for your time but I don't think this is right. That isssue was caused by a server error, and we should close that.

Also, I think that maybe you have committed over an old commit, because there are files (Utils.java) which have long been converted to Kotlin. Let me know if there's something I can help you with on Slack!

Issue #4832, opened 2022-09-04
Update the logo of open food facts in README.md


pratyaksh1610:
The logo of open food facts is not up to date.

VaiTon:
Yup, you're right! If you want to work on this I can assign you!

teolemon:
The logos of secondary projects aren't either.
https://drive.google.com/drive/u/1/folders/1-0BEVMzkGQ1bRZ6XkdTnX5h74ilkKRFD
https://drive.google.com/drive/u/1/folders/1IgeI4jtLulSLhKiK1nIy8RAsqW7y8DYA

teolemon:
We could update the logo in the app itself, and especially in its flavors

pratyaksh1610:
I think we can extend and update the issue involving two tasks :  

- [ ] Update the logo in README.md
- [ ] Update the logos (including 4 flavors Open Food Facts, Open Beauty Facts, Open Pet Food Facts and Open Products Facts in the android app

BhuvanAde:
As a UI/UX designer, I believe I can solve this issue as mentioned.

teolemon:
- Related: https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/4848

PR #4915, opened 2023-02-14
style: Added new logo in README.md

dmdbilal:
####  Description

Added new logo in README.md
#### Related issues

#4832


VaiTon:
We should still use the logo from the static website. Another alternative is to upload the logo to github. With this PR I don't know where is that link pointing

pratyaksh1610 (changes requested):
Also, please add a before and after image for easier review and for maintaining history.
Also the test fails. PTAL and send back for review.
Thanks.

dmdbilal:
@VaiTon I've downloaded the logos from the drive that was provided in the issue #4832. If that is not, then where can i get the logos ?

VaiTon:
@dmdbilal you should do as this does: https://raw.githubusercontent.com/openfoodfacts/.github/main/profile/README.md

dmdbilal:
@VaiTon Can you check now ?

PR #5028, opened 2023-07-08
fix: Fixed typo in german translation

hairy-tortoise:
####  Description
Fixed a typo in the german translation


VaiTon:
Hey, thank you for your PR! Unfortunately you should change the translation on crowdin[^1], otherwise it will be overridden on the next sync. 

[^1]: https://translate.openfoodfacts.org/

PR #5194, opened 2024-04-01
build(deps): Bump Truth to 1.4.2.

cpovirk:
This requires adapting to Truth's recent nullness annotations.

Specifically, a `Subject` constructor (and the associated
`Subject.Factory` and `assertThat` function) should accept a null actual
value. By doing so, it makes it possible to write
`assertThat(foo).isNull()`, for example.

This does not necessarily mean that all assertions will _pass_ if the caller
passes `null`. Obviously `isNotNull()` will fail, of course, but so too may
other assertions, like `hasFoundProducts` will here.

(In principle, it would be nice for all assertions to fail _with
detailed `NullPointerException` messages_. In practice, we mostly
haven't done that even for the core Truth assertions. So I haven't here,
either.)

Without this PR, the error that Truth 1.4.2 produces is:

```
ProductsAPITest.kt:245:46 Type mismatch: inferred type is KFunction2<FailureMetadata, Search, ProductsAPITest.SearchSubject> but (FailureMetadata!, TypeVariable(ActualT)?) -> TypeVariable(SubjectT)! was expected
```

Fixes https://github.com/openfoodfacts/openfoodfacts-androidapp/pull/5191

### DIFF
--- a/app/src/testOff/java/openfoodfacts/github/scrachx/openfood/network/ProductsAPITest.kt
+++ b/app/src/testOff/java/openfoodfacts/github/scrachx/openfood/network/ProductsAPITest.kt
@@ -228,23 +228,23 @@ class ProductsAPITest {
 
     class SearchSubject private constructor(
         failureMetadata: FailureMetadata,
-        private val actual: Search,
+        private val actual: Search?,
     ) : Subject(failureMetadata, actual) {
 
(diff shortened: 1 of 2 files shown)


teolemon:
we're not actively maintaining the old app. I'm going to merge this PR anyway, thanks @cpovirk 👍

PR #5267, opened 2024-08-06
fix: Fix ProductsAPITest

g123k:
Let's try to fix remaining tests: https://github.com/openfoodfacts/openfoodfacts-androidapp/actions/runs/10227597751/job/28299218293


teolemon (approved):
let's release

PR #5268, opened 2024-08-09
Update FUNDING.yml

Poussinou:
Hi,

This small PR adds your OpenCollective ID to the FUNDING.yml inside this repo. OpenCollective will now show up directly when clicking on the "SPONSOR " button. It should also be automatically collected by F-Droid and added to the information on the F-Droid website and inside the app.

Cheers!

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,4 +1,5 @@
 # Docs: https://docs.github.com/en/github/administering-a-repository/displaying-a-sponsor-button-in-your-repository
 
 github: openfoodfacts
+open_collective: OpenFoodFacts-Server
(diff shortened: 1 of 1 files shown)


teolemon (approved):
Great @Poussinou 👍 
Note that we also have a default FUNDING.yml for the whole org, so that we could almost delete this file: https://github.com/openfoodfacts/.github/edit/main/.github/FUNDING.yml

Poussinou:
> Note that we also have a default FUNDING.yml for the whole org, so that we could almost delete this file: https://github.com/openfoodfacts/.github/edit/main/.github/FUNDING.yml

Actually you're right, let's completely remove the one from this repo, so that the master one can be used.

Issue #4069, opened 2021-06-20
other_editors should include the placeholder directly in the string


teolemon:
other_editors should include the placeholder directly in the string. Otherwise, we'll have issues in RTL languages, or languages where the order of words changes.
```
name="creator_history" 
Product added on %1$s at %2$s CET by %3$s

name="last_editor_history"
Last edit of product page on %1$s at %2$s CEST by %3$s

name="other_editors"
Product page also edited by 
```

Rajdeep-Paul-117:
Hi,
Could you explain a little more like where should we look for it?

PR #5463, opened 2026-01-07
fix(i18n): use formatted placeholder for other_editors

priyanshusky0:
This PR moves the placeholder into the `other_editors` string resource and updates its usage to use proper formatted strings.

This fixes i18n issues for RTL and other languages where word order differs, as discussed in #4069.


priyanshusky0:
This PR addresses the i18n issue described in #4069 by moving the placeholder into the string resource and updating its usage to proper formatted strings.  
Let me know if any adjustments are needed.

PR #5558, opened 2026-09-11
docs: Revise warning to important notice about app status

hangy:
Updated warning section to an important notice about the new Open Food Facts app and the archived status of the repository.


teolemon:
@hangy I was thinking of unleashing Antigravity on it once I get my quota back on Tuesday, to get it into a somewhat working state (and possibly release a send-off update for F-Droid before we archive)

hangy:
I did try fixing the Gulp build locally before, but the major upgrade breaks some stuff and requires several changes and switching to other libs.

I didn't think the effort was worthy, considering the app hasn't been updated from this in years 🤷‍♂️

